PR #2, opened 2011-11-15
Nifty 1.3.1 - Slick Renderer

mkaring:
New and improved slick renderer implementation fitting Nifty 1.3.1 as promised.


void256:
Thank you! :)

nifty-controls/src/test/java/de/lessvoid/nifty/controls/SelectionCheck.java
```diff
@@ -24,7 +24,7 @@ public class SelectionCheck {
     assertEquals(selection.length, listBox.getSelectedIndices().size());
     int i = 0;
     for (int o : selection) {
-      assertEquals(o, listBox.getSelectedIndices().get(i));
+      assertEquals(o, listBox.getSelectedIndices().get(i).intValue());
```
void256, line 27: I'm not sure why you've changed that though. You don't like auto unboxing of Integer to int or so? =) Is there a reason to use .intValue() here that I'm not yet aware of?

I've already merged your pull request ... I'm just curious :)

mkaring, line 27: This showed an error for me when compiling against Java 7 and the latest JUnit. Showed that assertEquals was ambiguous for int, Integer... using intValue fixed it and it does not hurt older versions.

mkaring:
You are welcome ;) I hope you like it all in all ^^

nifty-controls/src/test/java/de/lessvoid/nifty/controls/SelectionCheck.java
void256, line 27: ah! makes sense then! thanks for pointing it out! :)

PR #5, opened 2011-12-20
Slick-Renderer: Added some additional methods to initialize nifty in the Slick-Renderer

mkaring:
Hi

I added some methods that come handy... came to the idea when writing the documentation.

Nitram


void256:
and thanks again! =)

PR #6, opened 2011-12-23
Slick-Renderer: Fixing font handling

mkaring:
Title says all. Its not a nice way to fix the sizing problem. But as default implementation it will do the job.

Sadly Nifty does not request the size of a font when loading it... would be handy for fonts that get scaled freely, such as TTF fonts.


void256:
Ok, merged but I don't understand the: "Sadly Nifty does not request the size of a font when loading it" part. What does that mean? :)

PR #14, opened 2012-03-24
Bugfix: nifty.resolutionChanged() not working using JoglRenderDevice

telamon:
The `JoglRenderDevice` only reads the viewport dimensions once, so calling the `resolutionChanged()` method on nifty when your viewport has changed does not reposition your elements correctly.

I solved this by adding a public `invalidateViewport()` method to `JoglRenderDevice` that clears cached values; 

Simply call `invalidateViewport()` before calling `resolutionChanged()` and everything works as expected

The only downside is that it's just another "gotcha" a JOGL user will have to face when he's introduced to nifty.
Maybe this could be implemented using the eventbus instead if resolutionChanged() generates an event.

Or to let  `JoglRenderDevice` implement `GLEventListener` and automatically clear values on `reshape` event.
That would also let us handle the `dispose` event, but i'm not sure that works with nifty at all cause the RenderImage interface does not support texture-recreation. 
I've had to handle `dispose` events in my earlier builds on Win64 but in the end i resolved to completely recreate the nifty instance when that happens.

### DIFF
--- a/nifty-renderer-jogl2/src/main/java/de/lessvoid/nifty/renderer/jogl/render/JoglRenderDevice.java
+++ b/nifty-renderer-jogl2/src/main/java/de/lessvoid/nifty/renderer/jogl/render/JoglRenderDevice.java
@@ -112,7 +112,7 @@ public class JoglRenderDevice implements RenderDevice {
         }
         return viewportHeight;
     }
-
+    
     private void getViewport() {
         final GL gl = GLContext.getCurrentGL();
         gl.glGetIntegerv(GL.GL_VIEWPORT, viewportBuffer);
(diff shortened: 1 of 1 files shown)


telamon:
please ignore the `dispose` subject it's unrelated to this issue.

void256:
In the past Nifty called getWidth() and getHeight() for the RenderDevice very often - each time the width and height of the current display mode was needed. I think this is why RenderDevice implementations cached the values.

With the introduction of the auto-scale feature Nifty will cache these values internally (in it's RenderEngine) and will only request the current values from the RenderDevice when some change happens (like someone calling resolutionChanged()).

Therefore I've modified other RenderDevice implementations to not cache these values anymore and only request the "current" values from GL. I've missed to modify the JoglRenderDevice since - at that time - it was not yet part of the new Nifty project structure.

I think it would be better to modify the JoglRenderDevice accordingly.

Would that work or is there something else that needs to be considered?

telamon:
Ah yes that would work perfectly, I just got into nifty a few days ago and i dared not strip the dimension fetching logic off its cache in case it was called with high frequency. I'll close this pull request , may make this change and submit a new pull, as I'm working on it now anyways?

PR #16, opened 2012-03-26
Feature: Element.getDescendants()

telamon:
Function added to Element.java:

```
 /**
 * get all children and all childrens' children.    
 * @return an iterator that will traverse the entire Element tree.
 */
 public Iterator<Element> getDescendants()
```

I've worked with many other node-tree'ish implementations and realized that getting a List or Iterator of all descendants of a node was missing in nifty.
I hope I haven't missed an existing interface somewhere that already accomplishes this.

I've created a simple tree traverser using the `Iterator` interface in order to avoid pre-generating lists of elements and waste resources.

The iterator implementation is called `ElementTreeTraverser` and I've put it in `de.lessvoid.nifty.elements.tools` package.

Feel free to accept this request if you think the feature is 'nifty' enough :)
Cheers!


void256:
Thanks! Looks very Nifty ;-)

Can you maybe add a Junit-Test - at least for ElementTreeTraverser?

telamon:
Oh ofcourse, sorry I missed the the entire testsuite for nifty, i'll be back.

PR #21, opened 2012-04-07
Event issue1 - 

telamon:
Hi!

I've taken the liberty to expose  `Element#canHandleMouseEvents()` as public, I'm not sure if it's a good idea
and i'd love some feedback on this move. 

Here's the story behind that decision:

I found a bug in my software yesterday, it was related to nifty-popup layers that remain after you've called `nifty.closePopup('id');` , At first i thought it was a bug, but then I noticed that they get marked as `done = true`.

The problem in question is my mouse event handling filter;
I've chosen not to rely on `Element#buildMouseOverElements()` because it adds the MouseEvents in question to the nifty event processing que, and that is not my intention.

So to avoid that, I'm manually iterating through all elements and check their bounds and confirm that they're capable of handling the mouse event before I let the InputSystem pass it on to nifty for processing.
But my problem is that i've been using `Element#isVisisbleToMouse()` method uptil now to detect if an element can handle mouse events, but that method reports `true` for an "done" popup-layer when I need a `false` in that case.

I could not see any way to reach the `done` boolean through the api, 
so as a last resort I exposed `Element#canHandleMouseEvents()` as public.

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/elements/Element.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/elements/Element.java
@@ -1340,7 +1340,7 @@ public class Element implements NiftyEvent<Void>, EffectManager.Notify {
    * Checks if this element can handle mouse events.
    * @return true can handle mouse events, false can't handle them
    */
-  boolean canHandleMouseEvents() {
+  public boolean canHandleMouseEvents() {
     if (isEffectActive(EffectEventId.onStartScreen)) {
       return false;
     }


void256:
I think changing `boolean canHandleMouseEvents()` to a public method might help in other situations as well. I've merged it. Thanks!

But :)

I think what we really need is something else. There is a `public String getMouseOverInfoString()` method on the Screen class. It returns a String that might be of interest for you :)

Nifty internally already knows the information you are trying to get it seems. The getMouseOverInfoString() method will be forwarded to the `MouseOverHandler` class that keeps that kind of information. The information itself is not public available yet but can you check if this is what you are after?

It would be possible to make that public I think as well :)

telamon:
Yes my first attempts at this event filter was using getMouseOverInfoString() , I don't remember exactly why i discarded that code but one reason was that I thought that building a string and then regex:ing it would add a bit too much overhead to my filter as i'm doing that calculation for every mouse event recieved. But everything works as intended right now without any noticeable slowdown.

Thanks for the pull! =D

void256:
I was pretty sure that you know about getMouseOverInfoString() and I agree that parsing this String would be silly :)

But what about a getMouseOverElements() that gives you a List of Elements (sorted front to back) that the mouse currently hovers? Nifty already keeps that information internally and it's not necessary to calculate this again by yourself.

Would that help your use case? I can add this pretty easily I think...=)

telamon:
Yes i was considering getMouseOverElements() at some point too. weirdly enough i can't find that method now. 
Unless my memory serves me correct it required a MouseEventHandler or a NiftyMouseEvent as parameter, and that defeated the purpose of my filter as i'm trying to prevent some events from ever reaching nifty. 

I don't think regular nifty usage would ever require a filter like this; I'm developing a digital-painting-application and when the user has begun a brush-stroke all mouse events must exclusively be forwarded to the painting logic even if the user's stroke travels over a nifty component, so that's why i need a filter that gives exclusive rights to one of the two event-receivers  depending on wether a mouse-down gesture starts on a nifty-control or on the canvas.

So I've managed to accomplish that flow by choking my `NiftyInputSystem` implementation, but all filter logic had to be calculated before events reached the nifty queue. 

Hope that sheds some light on why I'm doing things so awkwardly and not using existing methods. =)

PR #24, opened 2012-04-19
Updated style of tabs to make them look Nifty!

mkaring:
Title says everything. Now they look Nifty on a acceptable level.


void256:
great work with the transparent pixel line! :D

PR #27, opened 2012-05-16
Cleanup

toolforger:
Warning elimination:

Removed unused imports.

Commented out unused private members and locals.
I was unsure about SpecialValuesReplaceTest.MyObjectCallback.getValue, so I
just left it in and added a @SuppressWarning.

Changed EffectsType.translateSpecialValues so that it uses Arrays.asList
instead of a generic varargs parameter (which is a combination that the Java
compiler cannot properly check for type safety).

Code simplification:

Removed gratuitous parameter on private function
PaulsSoundSystemSoundDevice.addAdditionalCodecs.

Removed the unused generic type parameter on NiftyEvent and its subclasses.
This is a compatible change, except for applications that try to Do the Right
Thing and added a <?> parameter (which can simply be removed).
Note that generic implementors of NiftyEvent have stayed generic. See
MenuItemActivatedEvent for an example.

### DIFF
--- a/nifty-controls/.settings/org.eclipse.jdt.core.prefs
+++ b/nifty-controls/.settings/org.eclipse.jdt.core.prefs
@@ -1,4 +1,3 @@
-#Sat Jul 17 14:59:01 CEST 2010
 eclipse.preferences.version=1
 org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
 org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.6
@@ -9,5 +8,4 @@ org.eclipse.jdt.core.compiler.debug.localVariable=generate
 org.eclipse.jdt.core.compiler.debug.sourceFile=generate
 org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
 org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
-org.eclipse.jdt.core.compiler.problem.forbiddenReference=warning
 org.eclipse.jdt.core.compiler.source=1.6
--- a/nifty-controls/src/main/java/de/lessvoid/nifty/controls/ButtonClickedEvent.java
+++ b/nifty-controls/src/main/java/de/lessvoid/nifty/controls/ButtonClickedEvent.java
@@ -2,7 +2,7 @@ package de.lessvoid.nifty.controls;
 
 import de.lessvoid.nifty.NiftyEvent;
 
-public class ButtonClickedEvent implements NiftyEvent<Void> {
+public class ButtonClickedEvent implements NiftyEvent {
   private Button button;
 
   public ButtonClickedEvent(final Button button) {
--- a/nifty-controls/src/main/java/de/lessvoid/nifty/controls/ChatTextSendEvent.java
+++ b/nifty-controls/src/main/java/de/lessvoid/nifty/controls/ChatTextSendEvent.java
@@ -11,7 +11,7 @@ import de.lessvoid.nifty.controls.chatcontrol.ChatControl;
  *
  * @author ractoc
  */
-public class ChatTextSendEvent implements NiftyEvent<Void> {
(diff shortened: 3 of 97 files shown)


toolforger:
(Still using English since this is public.)

This is a preliminary cleanup before I start my real work - the work on the builder framework will require some use of generics and I needed all generics-related warnings out of the way to see whether I'm doing everything right.
Feel free to reject as appropriate.

In some cases, this merely replaced a "rawtypes" warning with an "unchecked" one, which had to be silenced with an annotation. However, "unchecked" is unavoidable when casting to a generic type (thanks to type erasure), so it's okay in these situations; "rawtypes" can mask avoidable errors.

I made DropDownNull and ListBoxNull generic, on the grounds that an empty List<Integer> still has a different type than an empty List<String>.
The alternative would be to make them nongeneric by filling the type parameter with yObject>.
It's a fine point and either decision may do more harm than good, it depends on how these classes are used, which I don't know. I'll change it to <Object> if you prefer that.

Some static imports were gone. If that's a problem, they can be restored as needed by doing a static "star import" of the class (i.e. pulling everything in from there), then doing Eclipse's "clean up imports" command. (I don't know how to achieve this in Netbeans.)

I did not address EasyMock-related "rawtypes" warnings since I didn't understand the isA function's semantics well enough. So I just silenced these warnings with annotations; I'll deal with the issue when I find I need EasyMock in new tests.

I tried to get rid of deprecation warnings, but it turned out it wasn't so easy to decide where the division line is between @Deprecated (deprecating code that deals with deprecated stuff and is itself legacy already) and @SuppressWarnings("deprecation") (perfectly valid code that calls into deprecated code as an implementation detail - that's actually a valid use if you wish to stay compatible with existing code that subclasses your code).
So I reverted these changes.

The @Deprecated annotations on XxxControl classes seem strange.
The comment indicates that they are just a way to discourage people from using them for application code, but the comment is a bit too terse to explain how to transform existing code; it's certainly not enough to rewrite the demo code that uses XxxControl (well, to me, but then I haven't used Nifty in anger yet, so maybe an actual user of Nifty would know what to do).

void256:
Well, thanks for all of your work BUT I can't apply your pull request with a clear conscience. Sorry!

First of all it's way too huge to be judged throughout for any side effects on existing code. And it's all in one giant single commit too. It might be better to group your changes into logical single commits. That would increase the chance that we'd be able to apply at least some of your changes.

I don't deny that you're changes are not correct or helpful. It's just that I can't judge them in their current form without checking each of the "242 additions and 263 deletions". And there might be cases where you don't have all the necessary information to make the correct decision. And btw: It's not like I could: Some of the code you've changed was contributed by other people and I'd like to have them to review your changes.

Last but not least I don't understand why you need to change stuff all over the place for your planned improvements of the Builder classes :) I mean, what has PaulsSoundSystemSoundDevice to do with that change? =)

I feel bad turning down these changes, but it’s what I have to do.

toolforger:
I needed to get rid of the generics-related warnings since the Builder work requires generics.
The other changes were those trivial things that were easier to fix than to silence via project-specific settings on a dozen projects.

I could split this one up. Just give me an example of an acceptable commit; I don't want to split into a gazillion changes if a handful would have sufficed.

void256:
Your new "Import cleanup" pull request has been applied now. It was much better since it more or less concentrated on a single thing that the patch fixed. Thanks!

About your other changes:
What about creating an issue here on github so that we can discuss things first? I mean not for simple things like your import cleanup but for more complicated things like removing generics from the NiftyEvent. Or for things that appear strange to you or things you are unsure about? You could even consider to create an issue for discussion first when things are totally obvious to you. Maybe there is/was a reason for it besides pure stupidity .. which does not imply that the original reason for things was not pure stupidity though ;D

toolforger:
Ah, I didn't know you use the issue tracker, as it is empty.
Erm.
All closed. Okay :-D

void256:
:)

PR #32, opened 2012-05-17
Implemented proper shutdown of effects after there lifetime.

mkaring:
As discussed in IRC. It all still working as it seems ;)


void256:
Nifty allows 120 characters per line. Thats ok. No reformat necessary!

We live in a world of 16:10 or 16:9 ... I can probably display lines of length > 200 chars without a problem here ;D

void256:
... when I push really hard I can display lines of 330 chars ;D

but on a more serious note:
I agree that too long lines are hard to read but really, 120 chars is not too long in my world

PR #34, opened 2012-07-02
Only display console cursor when the textfield has focus.

odyssomay:
The console cursor is currently always displayed, even if the textfield isn't in focus. 
The solution is simple (use the same custom effect as the original textfield).


void256:
Thanks :)

PR #35, opened 2012-07-04
Console border cut off

odyssomay:
There's a problem with the border in the default console style; the border is cut off when the scrollbar is activated.

<a href="http://tinypic.com?ref=idel3c" target="_blank">Image and video hosting by TinyPic</a>

The upper picture shows how it looks with this fix, the lower is the default style.


void256:
nice catch thanks!

PR #46, opened 2012-09-07
Dropdown control improvement

mkaring:
Added possibility to the dropdown control to set the converter class using the builder or the XML


void256:
nice! seems that was indeed missing :)

PR #57, opened 2012-10-09
Improved TextBox

mkaring:
- The initialization of the textbox was improved by removing a useless initialization of the text that is displayed.
- Furthermore a bug was removed that caused a text field with error contents to be corrupted once a screen was entered again
- Missing documentation was added

### DIFF
--- a/nifty-controls/src/main/java/de/lessvoid/nifty/controls/textfield/TextFieldControl.java
+++ b/nifty-controls/src/main/java/de/lessvoid/nifty/controls/textfield/TextFieldControl.java
@@ -66,8 +66,13 @@ public class TextFieldControl extends AbstractController implements TextField, T
     fromClickCursorPos = -1;
     toClickCursorPos = -1;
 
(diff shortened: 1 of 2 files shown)


nifty-controls/src/main/java/de/lessvoid/nifty/controls/textfield/TextFieldControl.java
```diff
@@ -113,7 +118,6 @@ private void activateFilter(final String filter) {
   public void init(final Properties parameter, final Attributes controlDefinitionAttributes) {
     focusHandler = screen.getFocusHandler();
 
-    textField.setTextAndNotify(textElement.getRenderer(TextRenderer.class).getOriginalText());
```
mkaring, line 116: This line caused textfields that use a display filter to be overwritten with the filtered content. This is problematic with password fields for example as the password is overwritten with the password replacement character.

After all this line is a left-over of the old text box and serves no real purpose anymore as the text does not need to be initialized again.

Issue #58, opened 2012-10-10
Can not load fonts in sub folders


Frdnspnzr:
In version 1.3.2 it's not possible to load fonts referenced like "./font/myfont.fnt". The source of this error is slick2d.render.font.loader.AngelCodeSlickRenderFontLoader.class (and possibly other loaders as well).

The slick renderer tries to extract the image filename from the font file, in the example that would be something along "myfont.png". This filename is used to load the Angel Code Font in Slick. However, it's wrong because it's relative to the font file, not the working directory. The right filename would be "./font/myfont.png".

PR #59, opened 2012-10-10
Use file path - Fix for issue #58

Frdnspnzr:
Angel Code Font Loader now takes paths into account and uses image
paths relative to the working directory, not relative to the font file.

### DIFF
--- a/nifty-renderer-slick/src/main/java/de/lessvoid/nifty/slick2d/render/font/loader/AngelCodeSlickRenderFontLoader.java
+++ b/nifty-renderer-slick/src/main/java/de/lessvoid/nifty/slick2d/render/font/loader/AngelCodeSlickRenderFontLoader.java
@@ -31,12 +31,16 @@ public final class AngelCodeSlickRenderFontLoader implements SlickRenderFontLoad
       /* Slick is unable to extract the image files from the font definition. */
       BufferedReader in = null;
       try {
+    	String imageFilepath;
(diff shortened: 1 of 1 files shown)


void256:
@Friedenspanzer 
I agree with toolforger.  We should fix this in an OS independent way. Do you plan to change it yourself or should I fix this?

Frdnspnzr:
Yes, you're right about OS independence, did not think of that, sorry.

@toolforger Now it's _relative_ to the working directory. Before it was _relative_ to the font files directory.

toolforger:
What's your definition of "working directory"? Of "font directory"?
I'm under the impression that we have different definitions, and are talking at cross-purpose here.

Frdnspnzr:
Also, the discussion has absolutely nothing to do with the problem at hand :)

The same as it is in Unix (see pwd): The directory we're currently working in, where Java will look for files. By default, the directory we've invoked the java command. I've overwritten this with something like "/Users/Me/myAwesomeProject", so I mean exactly this folder by working directory and relative to this my font is in "./font/font.fnt".

toolforger:
Then I don't understand why you're mentioning the working directory at all.
Your code is independent of whether the path it's manipulating is absolute or relative.

void256:
finally fixed. see this commit https://github.com/void256/nifty-gui/commit/cf1d1d4f861f98365d0c1805bb19b6cd831a831b

;-)

PR #62, opened 2012-11-11
Tree Control

mkaring:
I created a new version of the tree control that appears to be working nicer then the original one.
If you have any additional proposals on how to improve it, I'd be glad to hear them.


void256:
Great work! Is it really necessary for the ListBoxItemProcessor to be part of the public API? In what cases does user code need to provide their own implementation?

mkaring:
For example the TreeBox requires it ;-)
For the rest... don't know. I will start to implement the TreeBox into Illarion during the next days. I guess I'll get a better overview over what is missing and what is still needed then.

PR #72, opened 2012-12-14
Optimized element building performance

mkaring:
Applied some optimizations (by removing screen.layoutLayers) to the core of Nifty and to the ListBox control


nifty-controls/src/main/java/de/lessvoid/nifty/controls/listbox/ListBoxControl.java
```diff
@@ -538,8 +538,6 @@ private void initializeScrollPanel(final Screen screen) {
     childRootElement.setConstraintY(new SizeValue("0px"));
 
     updateBottomRightElement();
-    nifty.executeEndOfFrameElementActions();
```
void256, line 541: Did that hurt performance that much? I think this might be needed in some circumstances. Although I can't remember it exactly BUT these lines look like they would break some stuff eventually when they are missing :)

Everything else looks pretty good! Can I get these two lines back eventually?

mkaring:
As you requested.

void256:
excellent work! thanks!

PR #77, opened 2013-01-04
Removed useless listener notification during layout

mkaring:


void256:
"probably" useless listener notfication would have been the better wording ;-) but it really looks useless

PR #89, opened 2013-01-27
1.4 Pull request: bugfix, getters for padding and margin

fleather:
- Critical bugfix regarding my change of commit 2bf94cba63, Element.findElementById() didn't work correctly
- Changed method Screen.findElementById() according to Element.findElementById()
- Getter methods for Element padding and margin


void256:
Thanks for fixing this :)

In general I'd prefer if you could organise the pull requests so that an individual pull request contains a single change. It could contain multiple commits of course but only if they are related to the same issue.

This would make the review a lot easier. Otherwise it would require cherry-picking some commits in case some of the changes would be rejected.

That's not the case in here so that's fine for now tho.

PR #90, opened 2013-01-28
Added map to Element for custom user data

fleather:
for example game entity objects.


void256:
I like the generic method better ;-) do we need to keep the none-generic one as well? Why? Element is kinda a core class and I'd prefer if we could remove the none-generic methods. Would this be possible or are these methods really necessary for some special case that I can't think of?

I just want to make sure that you haven't left the old methods in their by accident! ;-)

void256:
I'M STUPID! Commit is fine and now applied - sorry for the trouble :-)

PR #91, opened 2013-02-06
insert a child Element in a specific position in parent elements list

relu91:
Added methods to insert a child Element in a specific index of parente elments list . So you can also insert a child element in a Horizontal layout or Vertical layout in your position and not only at the last as was before.

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/Nifty.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/Nifty.java
@@ -1416,19 +1416,28 @@ public class Nifty {
       return ReferenceStrength.STRONG;
     }
   }
-
+  /**
+   * Creates an element from a type at index position in Parent child list
+   * 
+   */
+  public Element createElementFromType(final Screen screen, final Element parent, final ElementType type,final int index) {
+	    if (type instanceof LayerType) {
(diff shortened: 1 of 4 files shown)


void256:
Thanks for this Christiano! This looks actually very well.

Unfortunatly I can't apply it directly because I'd like to keep the 1.3 branch clear of new features - except very very few special cases like the batched renderer I'm working on right now. Although I don't think your changes will break anything I'd appreciate if you could make the same changes on the 1.4 branch and send another pull request.

Besides that something changed every line in ElementBuilder.java as well. I guess this is a CR/LF issue which maybe your IDE or editor has changed automatically. This makes seeing your actual changes quite impossible :) at least without any additional effort. Can you check this and please make sure that only lines you've actually changed show up in the diff?

If there is some general issue with CR/LF in this file we can modify that later in a separate commit but for this change I'd like to be able to see the difference :)

relu91:
Ok i will make an other pull request in 1.4 brach! I didn't know which one was right and i chose the wrong one! :D . I'm sorry for ElementBuilder in the next pull request i'm going to check all the diff files before to commit! I already tested this in my editor and it works very well :) it also allow to an user to modify order in horizontal and vertical layout that is quite cool! :) See you in the next pull !

void256:
Awesome! Thank you :)

I'll close this pull request now.

PR #118, opened 2013-05-02
Color field added to image element.

gsimard:
A color field was added to image elements so a white image can be drawn with a color without resorting to overlay effects.


void256:
nice work! looks great! :smile:

PR #125, opened 2013-06-28
JogAmp 2.0.2-rc12 update

xranby:
nifty-renderer-jogl2/pom.xml: Update jogamp.version to 2.0.2-rc12

Signed-off-by: Xerxes Rånby xerxes@zafena.se


void256:
Excellent ;)

PR #147, opened 2013-09-26
Various minor fixes.

ghost:
Hi again, @void256. I separated these changes out of the LibGDX stuff because they are unrelated minor improvements to Nifty that I came across as I was developing the LibGDX renderer. A couple of the changes will also improve the Nifty LibGDX desktop examples. Enjoy! :smile: 

Details:
- Add compile scope to xpp3 dependency in core pom. The scope was
  missing, and the default is 'compile', anyway. It's best not to rely
  on the defaults.
- Fix spelling errors in NiftyInputConsumer and ControlsDemo.
- Remove javadoc for nonexistent parameter in BatchRenderBackend.
- Add @Override to toString method in ResolutionControlJOGL.
- Fix getScreen method in ControlsDemo.  The getScreen method returns
  "demo" when it should return "start". The "demo" screen is the second
  screen in the ControlsDemo example. If someone were to use this method,
  they would miss the intro on the "start" screen.  Plus, all the
  getScreen method in every other NiftyExample implementation return
  "start".
- Move comment to the correct location. Move the "internal
  implementations" divider comment above the first private method in
  LwjglBatchRenderBackend, where it should be. There is one private method
  that is above the comment.
- Fix parsing of the "fillRemovedTexture" property. In the LWJGL and JOGL
  batch render backends (both core and non-core), fix the parsing of the
  boolean system property called "fillRemovedTexture". The current
  implementation doesn't respect the default setting; it will always be
  false, even if the default setting is set to true. The problem is that
  it should use the parseBoolean method instead of the getBoolean method
  because getBoolean tries to call System.getProperty, when
  System.getProperty is already being called, so it would result in a
  redundant call to System.getProperty, whereas the parseBoolean method
  just simply converts the result of System.getProperty from a String to a
  Boolean, if the value is "true" or "false", which is exactly what we
  want.
- Fix crashing bug in LwjglBatchRenderBackend. Don't attempt to render an
  empty batch, i.e., if its vertex buffer is empty, because the render
  method will attempt to access non-accessible elements, throwing some
  nasty exception and crashing the program. This happens when the batch
  pool reaches maximum capacity, causing it to double in size, i.e., from
  2 to 4, and the problem is that in some cases it will only have enough
  data for the 3rd batch, and the 4th batch will be completely empty.
- Fix example with dynamic drop down list. The problem is that there is
  not enough room for the dynamic drop down list to be inserted above the
  static drop down list in the drop down example, so increase the size of
  the spacer panel between them from 8px to 70px to make room. Otherwise
  they overlap, which looks bad. This may be a bug in Nifty, as it seems
  like it should make space on the fly instead of having to manually do
  it.

### DIFF
--- a/nifty-core/pom.xml
+++ b/nifty-core/pom.xml
@@ -27,6 +27,7 @@
       <groupId>xpp3</groupId>
       <artifactId>xpp3</artifactId>
       <version>1.1.4c</version>
+      <scope>compile</scope>
     </dependency>
     <dependency>
       <groupId>org.bushe</groupId>
--- a/nifty-core/src/main/java/de/lessvoid/nifty/NiftyInputConsumer.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/NiftyInputConsumer.java
@@ -22,7 +22,7 @@ public interface NiftyInputConsumer {
   /**
    * Process the given keyboard event.
    * @param keyEvent the keyboard event to process
-   * @return true, when the mouse event has been processed and false, if not
+   * @return true, when the keyboard event has been processed and false, if not
    */
   boolean processKeyboardEvent(KeyboardInputEvent keyEvent);
 }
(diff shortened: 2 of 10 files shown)


void256:
Thanks!

PR #149, opened 2013-10-12
Fix hover effects after screen transitions.

ghost:
Process a fake mouse event with current coordinates so that Nifty will
activate hover effects for any element currently under the mouse, and
will also deactivate any hover effects for any element NOT currently
under the mouse. This prevents the undesirable behavior of having an
element under the mouse that should be in a hovered-effect state, but is
not due to the screen change, and of having an element that is NOT under
the mouse that should NOT be in a hovered-effect state, but IS due to
the screen change.


ghost:
I'm going to resubmit this as I didn't realize there are some test failures I have to deal with.

PR #152, opened 2013-10-23
Fix off-by-one pixel bugs.

ghost:
This is really two separate but related bugs, related because they both
cause off-by-one pixel errors. The first affects nifty-core, but the
second only affects nifty-renderer-libgdx.

Bug 1) Many classes in the effects and layout packages in nifty-core
are using a method in the SizeValue class called getValue that returns a
floating point value. The issue is not with this method itself per se,
but with callers of the method, who really need an integer value, and
cast the returned floating point value to an integer, effectively
discarding everything to the right of the decimal. It can easily be
verified that calling getValue and then casting to an integer returns
incorrect values much of the time. This bug is now verified with a
simple test in SizeValueTest. The bug rears its ugly head the worst when
converting a simple pixel value of say "212px" from a SizeValue whose
percent is set to 100, to a floating point pixel value, and then casting
it to an integer: the result will always be 211 and NOT 212. There are
countless other values that return a similar result when the SizeValue
percent is set to 100.

The solution to this error, which rectifies issues such as creating
quads at off-by-one pixel sizes, is to change all callers of getValue
that actually need an integer value to use the preexisting getValueAsInt
method, BUT since it ALSO simply casts the floating point value of the
getValue method to an integer, persisting the bug, we modify the
getValueAsInt method to use Math.round, and round the result returned by
getValue to the nearest integer instead of casting it. This returns much
more correct values, and fixes the bug. This solution has been
integration-tested without noticing the introduction of any other
regressions.

There is an additional test in SizeValueTest that verifies that this
off-by-one SizeValue bug is fixed. The javadoc for SizeValueTest has
also been improved, especially noteworthy is the warning to NOT cast the
result of getValue to an integer, but rather to use getValueAsInt.

Classes that call SizeValue.getValue method, and do NOT cast to an
integer, but store the result in a floating point variable have not been
modified as this is a legitimate use of the getValue method.

Bug 2) nifty-renderer-libgdx is not configured to use exact openGL
pixelization. Configure nifty-renderer-libgdx to use exact openGL
pixelization for 2D rendering as officially recommended in a faq on
opengl.org at:

http://www.opengl.org/archives/resources/faq/technical/transformations.htm#tran0030

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/effects/impl/Border.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/effects/impl/Border.java
@@ -115,7 +115,7 @@ public class Border implements EffectImpl {
   }
 
   private int getBorder(final Element element, final SizeValue sizeValue) {
-    int left = (int) sizeValue.getValue(element.getParent().getWidth());
+    int left = sizeValue.getValueAsInt(element.getParent().getWidth());
     if (left == -1) {
       left = 0;
     }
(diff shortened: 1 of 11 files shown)


void256:
Great catch! Thanks again! :smile:

PR #153, opened 2013-10-23
Cleanup Element class.

ghost:
The Element class in nifty-core is badly in need of some basic cleanup.

Remove useless, tautological comments/javadoc. Add some useful
comments/javadoc where necessary. Improve comment/javadoc grammar,
sentence structure, and spelling.

Rename getElements to getChildren, add to addChild, and the overloaded
add method that takes an index parameter to insertChild because those
methods names are infinitely clearer. Rename these methods in all
calling classes. Also rename elements field to children.

Fix code style violations so that it matches the code style of the rest
of the project.

No actual logic is altered. All tests are passing, and
integration-testing reveals no problems.

### DIFF
--- a/nifty-controls/src/main/java/de/lessvoid/nifty/controls/MessageBox.java
+++ b/nifty-controls/src/main/java/de/lessvoid/nifty/controls/MessageBox.java
@@ -181,7 +181,7 @@ public class MessageBox extends AbstractController {
     }
 	
 	private void clearButtons() {
-		List<Element> buttons = messageboxPopup.findElementByName("#messagebox").findElementByName("#buttons").getElements();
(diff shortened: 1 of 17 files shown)


void256:
Excellent cleanup work again, Aaron! :smile: 

And I agree with your renaming of the core child Element managment methods too. The new names are a lot cleaner.

However, I have some concerns because this change will most likely break quite a bit of existing user code that relies on the old methods (f.i. custom controls).

Breaking existing code _is_ allowed in the 1.4 branch but I think we could make migrating to 1.4 a lot easier by keeping the existing methods and simply mark them as @deprecated for the time being.

Would that be possible? :/

ghost:
Done! :smile:

void256:
Excellent! :smile:

PR #158, opened 2013-10-24
Add displayItems method to DropDownBuilder.

ghost:
This is really a very basic feature that is missing for some reason. I set it up so that you can use an integer for the number of display items, which I think is much more elegant than forcing the user to use a String, either by having to put the integer in quotes, or, in my particular case, having to call String.valueOf on a predefined static constant.


void256:
This issue was mentioned on the jme3 forum here: http://hub.jmonkeyengine.org/forum/topic/nifty-changing-the-default-displayitems-size-of-a-drop-down-list/#post-246854

ghost:
Yep, I saw that before. ;)

void256:
Good job! :)

PR #172, opened 2013-11-23
Remove internal use of deprecated Element methods.

ghost:
Remove all uses of all deprecated methods in Element class. The
deprecated methods are still there, but they are not used anymore
internally.

There is also one deprecated method in Element class that is supposed to
delegate to the non-deprecated version, but instead it is duplicating
the non-deprecated version's code. Make this deprecated method delegate
to the non-deprecated version.

Also remove a duplicated word from a javadoc comment in the
getValueAsInt method of SizeValue class.

There are some faux changes present where the entire file appears to
have changed, but it's just due to the removal of 
 carriage return
characters in the line endings.

### DIFF
--- a/nifty-controls/src/main/java/de/lessvoid/nifty/controls/MessageBox.java
+++ b/nifty-controls/src/main/java/de/lessvoid/nifty/controls/MessageBox.java
@@ -138,10 +138,10 @@ public class MessageBox extends AbstractController {
 	}
 
 	private void setupMessageBox() {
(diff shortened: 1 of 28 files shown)


void256:
Excellent! :smile: I was thinking about doing the same but was too lazy =) Thanks for taking the time to clean that up!

ghost:
> void256 commented 23 minutes ago
> Excellent!  I was thinking about doing the same but was too lazy =) Thanks for taking the time to clean that up!

No problem!

PR #192, opened 2014-01-06
Changes in nifty (and all it's batch renderers) for new approach of rendering fonts; with separate texture for each font.

iamtakingiteasy:
Big conversion; to use multiple textures in font rendering for faster rendering: from 20-50 fps (due to big amount of sub-texture data uploading) to 1500-2000+ (by only texture-id switching) on my machine. Requires changes in jglfont to be applied, built and placed into local maven repo first in order to successfully finish the build process, since font renderer interface has been changed. Something in Gdx/JOGL batch renderers might have been accidentally broken during making necessary changes, needs extensive testing.


iamtakingiteasy:
https://github.com/void256/jglfont/pull/3 -- required pull into jglfont

iamtakingiteasy:
Initally, we have a blank texture (fig 1), backed up by ByteBuffer with some size and no data in it and allocated on GPU texture with the same dimensions.
Initial blank texture:

```
+--+
|  |
|  |
+--+
(fig 1)
```

We upload some glyphs (fig 2) BOTH to GL texture and CPU-hosted ByteBuffer
Uploading some glyphs:

```
+--+
|AB|
|CD|
+--+
(fig 2)
```

When there is no space for a new glyph, we dispose GPU-sided texture completly and creating a new one from ByteBuffer we had and allocate new texture with four times bigger size than previous one and initialize it with data from out ByteBuffer (fig 3).
Whoops, no more space for new glyph, re-sizing the texture, copying old texture data at 0,0 location of new texture:

```
         +----+
+--+     |AB  |
|AB|  -\ |CD  | 
|CD|  -/ |    |
+--+     |    |
         +----+
(fig 3)
```

Adding new glyphs again:

```
+----+
|ABEF|
|CDGH| 
|    |
|    |
+----+
(fig 4)
```

Adding even more glyphs:

```
+----+
|ABEF|
|CDGH| 
|IJKL|
|MNOP|
+----+
(fig 5)
```

Whoops, again no more space for a new glyph, re-sizing again, preserving old data:

```
          +--------+
          |ABEF    |
+----+    |CDGH    |
|ABEF|    |IJKL    |
|CDGH| -\ |MNOP    |
|IJKL| -/ |        |
|MNOP|    |        |
+----+    |        |
          |        |
          +--------+
(fig 6)
```

void256:
Thanks! Makes sense.

However some questions:
1. What if we run out of texture space? We need so many CJK font pages that we'd hit the texture size limit of the GPU in the next step? Lets say we have a 2k limit and we'll need 4k in the next resize step which is not supported anymore?
2. Why not start with a bigger texture and let the texture atlas handle everything?
3. With the BatchRendererTexture you introduced we already have a way to handle batches using multiple textures. Couldn't we just use that mechanism to quickly switch between multiple texture atlases each having a certain large size so that we don't need to "resize" at all? We just put your glyph pages into the first texture atlas (using the textureAtlasGenerator) until its full and then simply allocate another one and so on making sure that when we render the batches the correct textures are switches if necessary? The worst thing that could happen would be that we render large amount of texts which each render batch requiring a page from another texture atlas so that we would continuously switch textures all the time. But this might not soooo bad.

I'm typing this out of a dream since I'm already sleeping :smile: my appologies... I'll take a closer look tomorow...

ghost:
@void256 That's exactly what I'm thinking and this code makes me very
nervous.

I have multiple texture atlases (as well as non-atlas textures) already
implemented, but haven't had time to finalize it due to working on my own
projects. Absolutely nothing personal against the OP here, but if this gets
committed, it's going to significantly obfuscate the way batch rendering
works in Nifty, and I can probably forget about adding my code to Nifty
because the amount of work required to figure out how to get it to work
with this stuff will probably make it not worth it. I've put a lot of work
into it, and it might even solve all the issues discussed here. I'm also
concerned about being able to support the libGDX side of things with these
changes.

Aaron Mahan
Founder & Lead Developer
Forerunner Games, LLC
http://forerunnergames.com/

void256:
@3xp0n3nt I've already talked to @iamtakingiteasy on irc since I had some concerns with some of his changes too. I'll look into it a bit more tommorow and will talk with Alexander more. I'm sure we'll find a solution.

If you could share the work or ideas you've done so far it would be greatly appreciated as well. Bringing the work of both of you together would be pretty good, I think!

ghost:
@void256 Thanks, I'm working right now actually on getting my work ready for a pull request. It might be a day or two though.

iamtakingiteasy:
Okay, so i am closing this pull request and waiting for @3xp0n3nt work to base multi-atlas font rendering on.

PR #194, opened 2014-01-13
Implementation of setMousePointer for JoglInputSystem

bgroenks96:
I implemented setMousePointer for the JOGL input system using an instance of a NEWT Window now required in the constructor.

I also patched the associated demo JOGLNiftyRunner to work properly with this change.

Building full repository via Maven after changes was successful.


void256:
Looks good! Thanks! :smile:

PR #195, opened 2014-01-14
Increased visibility of mouse related nifty-renderer-slick input events

AODtorusan:
Hi

I am writing an alternative system to handle the `InputEvent`'s generated from Nifty/Slick2d, and found that the mouse related events have fields that are not visible from the outside. This makes it impossible to create a custom `AbstractSlickInputSystem`, as you cannot access the information contained in each event.

Would it be possible to increase the visibility of these fields (and add getters where necessary) ? 

Simon

### DIFF
--- a/nifty-renderer-slick/src/main/java/de/lessvoid/nifty/slick2d/input/events/AbstractMouseEvent.java
+++ b/nifty-renderer-slick/src/main/java/de/lessvoid/nifty/slick2d/input/events/AbstractMouseEvent.java
@@ -45,7 +45,7 @@ public abstract class AbstractMouseEvent implements InputEvent {
    *
    * @return the x coordinate of the event location
    */
-  protected final int getX() {
+  public final int getX() {
     return locX;
   }
 
@@ -54,7 +54,7 @@ public abstract class AbstractMouseEvent implements InputEvent {
    *
    * @return the y coordinate of the event location
    */
-  protected final int getY() {
+  public final int getY() {
     return locY;
   }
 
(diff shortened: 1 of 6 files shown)


void256:
I've talked to @mkaring (aka Nitram, on freenode irc in #niftygui) who is the original author of the Slick2d renderer and that was what he said:

```
[22:15:07] <void256> Nitram__: would be nice if you could take a look at that pull request - you probably had good reason to make those methods protected, I guess?
[22:16:24] <Nitram__> Yes I did
[22:16:45] <Nitram__> accessing the variables of these events is only meant to happen from the execution context of the event
[22:17:28] <Nitram__> you have to write a extension to this class anyway to get into the execution context and once you are there, you have access to the protected methods
[22:17:43] <Nitram__> there is absolutely no reason to expose those values further
[22:18:08]   <Nitram__> other than to promote a flawed application design
[22:21:27] <void256> do you want to explain that in a pull request comment or should I copy the chat log into the pull request?
[22:25:47]   <Nitram__> feel free to copy itI got to leave now
```

Does that make sense to you?

I tend to agree with him although it seems you have changed some more things besides making these methods public!? Do you want to explain that and especially why you're creating a custom SlickInputSystem?

I'm sure we can convince @mkaring if somethings missing :smile:

AODtorusan:
I find it simpler to directly process event objects from the event bus rather than the via the nifty `NiftyInputConsumer` or slick's `InputListener` interface. Similar to the `PlainSlickInputSystem` but with a non-empty `handleInputEvent(InputEvent)` method. I do this so I can use scala's pattern matching on the event objects and decide what to do accordingly.

I still think it's odd that the details for an event are private (or protected), and the full details can only be released to a `NiftyInputConsumer` or `InputListener`. It's analog in the awt system (`java.awt.event.InputEvent`) also has public accessors for all its details.

I indeed changed more than some access modifiers. I also added the required getters methods for some fields as these were missing altogether for some events.

PR #198, opened 2014-01-14
Fix annoying worthless dropdown warning.

ghost:
Clicking on a dropdown control generates a worthless warning, due to the
fact that it is using a popup control for the list. For some reason it
generates this warning message:

log.warning ("popup [" + id + "] already scheduled to be closed. Additional close call ignored.");

Change the warning to a fine log message since it's not a problem, and
it's a very detailed debugging message that no one cares about.


ghost:
No "epic!" shouts? :'(

PR #199, opened 2014-01-17
Fix JOGL batch renderer OpenGL initialization bug.

ghost:
Note: Also, remove unused logger.

Make GL2 instance static in JoglBatchRenderBackend and give it a static
initializer so that it will be initialized early enough to avoid a
NullPointerException during OpenGL initialization.

The GL2 instance has been being initialized in the constructor. However,
when it's parent class, OpenGlBatchRenderBackend, initializes its
constructor before JoglBatchRenderBackend's constructor, because
OpenGlBatchRenderBackend makes some openGL initialization calls in its
constructor (bad idea probably?), then a NullPointerException occurs
when the JoglBatchRenderBackend implementations of these openGL
initialization methods are called before the GL2 instance has even been
initialized in the JoglBatchRenderBackend constructor.


ghost:
Closing, as https://github.com/void256/nifty-gui/pull/201 makes this obsolete.

PR #200, opened 2014-01-17
Refactor image loading system.

ghost:
Note: All modules compile, all tests are passing, and all modules'
examples (LWJGL, JOGL, LibGDX) render correctly on my machine.

ImageLoader interface:
- Remove unused load methods.
- Refactor remaining load methods so that they make sense:
  - loadAsByteBufferRGBA & loadAsByteBufferARGB are the only needed options
  - loadAsByteBufferARGB has an option to flip the image vertically
  - get rid of "mouseCursor" in the interface names - image loading
    shouldn't need to know about mouse cursors (Single Responsibility
    Principle)
  - loadByteBufferARGB is used to load mouse cursors, setting
    shouldFlipVertically parameter to true
- Clean up code formatting, add missing final keywords to parameters,
  update JSR annotations.
- Rename image info get\* methods to increase clarity.
- In general, minimize the interface to provide only the functionality
  needed by the library to avoid bugs and confusion.

ImageIOImageLoader:
- loadAsByteBufferRGBA method adds alpha to all images instead of
  requiring the caller to try and decipher a long list of load methods
  that may or may not add alpha to the image. So all images will be loaded
  as RGBA.
- Rename to DefaultImageLoader, since it's used as the default image
  loader, and it doesn't need to expose implementation details in the
  name, which in a way violates encapsulation.
- Refactor the private methods, some of which were very long and
  complicated, splitting them up into small methods.
- Remove features completely unused (and unlikely to ever be used) by
  Nifty Gui such as passing around a transparency integer array
  (although there are others as well).
- Get rid of worthless comments and broken javadoc.
- Create an ImageProperties nested class to keep track of a loaded
  image's properties, which helped clean up the class tremendously and
  removes potential for hidden bugs - for example, if one of the image
  info get\* interface methods is called (such as getImageWidth), and is
  not set, it will now throw an exception instead of failing silently.
- Clean up code formatting, add missing final keywords to parameters,
  add missing JSR annotations. Adding final to some parameters revealed
  some areas where the parameter values were being modified.
- Update class to conform to new ImageLoader interface.

TGAImageLoader:
- This class has methods badly in need of refactoring, that have not yet
  been refactored. Add TODO comments where refactoring is needed.
- Get rid of worthless comments and broken javadoc.
- Clean up code formatting, add missing final keywords to parameters,
  add missing JSR annotations. Adding final to some parameters revealed
  some areas where the parameter values were being modified.
- Update class to conform to new ImageLoader interface.

Other Changes:
- Update all client classes to use the correct form of the new
  ImageLoader interface methods.


ghost:
Closing for now until I work out the merge conflicts with #201, and there's also a bug I need to fix in this PR that makes custom LWJGL mouse cursors invisible in the examples.

PR #201, opened 2014-01-18
Major batch rendering system overhaul, 2nd draft.

ghost:
- Switched everything from inheritance to composition & delegation.
- CoreProfile rendering system still not updated yet.
- LWJGL, JOGL, & LibGDX batch rendering systems are functioning correctly.

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/batch/BatchRenderImage.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/batch/BatchRenderImage.java
@@ -203,7 +203,7 @@ public class BatchRenderImage implements RenderImage {
   }
 
   private static void registerTextureSize(final int textureId, final int textureWidth, final int textureHeight) {
-    BatchRenderImage.textureSizes.put(textureId, new TextureSize(textureWidth, textureHeight));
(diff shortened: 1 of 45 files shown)


ghost:
@void256 This is a huge improvement over the previous overhaul ;). There's a bona fide GL interface to pass around, no more inheritance, no more protected methods, no more abstract classes, and yet really still no code duplication. The BatchRenderBackend implementations pretty much just delegate to OpenGLBatchRenderBackend. There's a lot more componentization, various small factory interfaces for creating the implementation-specific images, mouse cursors, buffers, etc, even factories for creating the BatchRendererBackend implementations themselves, since they're a bit too complex to require users to instantiate.

All they have to do now is `new BatchRenderDevice(LwjglBatchRenderBackendFactory.create());`, and it's the same for every backend. Check out the code after pulling; I think it's pretty...well, nifty. :D

ghost:
P.S. The BatchRenderBackend implementations' constructors are package private, so users MUST use the factory method - this helps prevent any mistakes.

ghost:
By "CoreProfile rendering system still not updated yet.", I mean that it doesn't use the new system yet, but **it still works**. It doesn't handle multiple texture atlases or non-atlas textures yet.

PR #203, opened 2014-01-22
Major batch rendering system overhaul, 3rd draft.

ghost:
Core profile batch rendering system:
- Refactor the entire core profile batch rendering system to match the
  new non-core profile batch rendering system setup / refactoring.
- Core profile now has multiple atlas & non-atlas texture support.
- Move all Core\* classes into nifty-core.
- Remove Core\* duplication between nifty-renderer-lwjgl &
  nifty-renderer-jogl.
- Abstract core profile OpenGL calls into CoreGL interface, which
  extends GL interface.
- Remove obsolete compatibility code / checking in Core\* classes, and
  require OpenGL 3.2 or higher in order to use the Core Profile
  rendering system at all. Since we requiring OpenGL 3.2 or higher, we
  don't need to pollute the system with all kinds of non-core profile
  compatibility junk.
- Remove rare uses of OpenGL extensions that are preventing the
  abstraction of CoreGL, since not all implementations can be forced to
  deal with extension-specific OpenGL stuff. It's not needed anyway.
  There are only several places where extensions are used, and they aren't
  anything critical. It limited the functionality slightly, but it
  shouldn't be missed. The badly-needed refactoring of the core profile
  system is so much more important anyway.
- Give *BatchRenderBackendCoreProfile classes package-private
  constructors to force users to rely on
  *BatchRenderBackendCoreProfileFactory.create() to instantiate. The
  reason being that the *BatchRenderBackendCoreProfile constructors are
  too complex and are managed automatically by their respective
  factories.
- Refactor any examples that are using a *BatchRenderBackendCoreProfile
  to use the factory method instead.
- Create JOGL & LWJGL implementations of CoreGL interface.
- Rename CoreCheckGL & CoreGLException to CheckGL & GLException,
  respectively, because there is nothing "core profile" about them, and
  the non-core profile system could easily use them if needed.
  LwjglInitHelper class in nifty-examples currently uses these in non-core
  profile code already. Move CHeckGL & GLException out of the core
  package and into the parent package, "batch", since they are not
  core-profile specific.
- Create CoreMatrix4f, CoreVector4f, & CoreMatrixFactory, based on Mat4,
  Vec4, & CoreMatrixFactory from nifty-renderer-jogl, for use in
  BatchRenderBackendCoreProfileInternal in nifty-core.

Non-core batch rendering system:
- Rename OpenGLBatchRenderBackend to BatchRenderBackendInternal because
  that is a more accurate name.
- Rename OpenGLBatch to BatchInternal because that is a more accurate
  name.
- Rename OpenGLBatchFactory to BatchFactory, in line with the last two
  mentioned changes.
- Add many non-core profile OpenGL methods that are only currently being
  used by the core profile system to the GL interface, rather than
  adding them to the
- General code cleanup.

Miscellaneous:
- Improve spelling, grammar, readability, formatting, & punctuation of
  existing javadoc and comments.
- Add missing final keyword to method parameters.
- Add missing JSR annotations.
- Fix incorrect JSR annotations.
- Add some asserts to deal with unsolvable JSR annotation warnings.
- General code cleanup

Caveats:
- nifty-renderer-libgdx does not yet have a core profile rendering
  system, which means mobile devices / OpenGL ES2 is not officially
  supported yet.
- I cannot integration-test the core profile rendering system because my
  computer is too old to support OpenGL 3.2, so I don't know if
  everything actually works. There may be (read: are probably) some bugs
  ;).
- There is still a lot of duplicated code between
  BatchRenderBackendInternal & BatchRenderBackendCoreProfileInternal.
- There are a few non-core profile OpenGL methods in the CoreGL
  interface because libGDX only supports them as part of the core
  profile (OpenGL ES 2 and up), so the nifty-renderer-libgdx non-core
  profile rendering system wouldn't have been able to implement them as
  part of the GL interface.
- There are major API changes.
- The Core\* classes look like every line changed because they of being
  moved to a different module within the project.


void256:
Alright! That looks great again. I kinda like the GL interface to remove code duplication! :smile: 

Getting the Core\* classes into Nifty core was not meant to be like that since there is that seperate CoreGL project for this. However seeing your GL and CoreGL interfaces we should really consider moving that abstraction into CoreGL as well to make things simpler there as well :smile: But that's something you didn't know and it can always be changed and moved to that separate project so that didn't hurt at all.

I've tried running the ControlsDemo with useCoreProfile = true and it crashes for me with this Exception:

`Exception in thread "main" java.lang.NullPointerException
    at de.lessvoid.nifty.batch.core.BatchRenderBackendCoreProfileInternal.bindAtlasTexture(BatchRenderBackendCoreProfileInternal.java:451)
    at de.lessvoid.nifty.batch.core.BatchRenderBackendCoreProfileInternal.updateAtlasTextureSection(BatchRenderBackendCoreProfileInternal.java:435)
    at de.lessvoid.nifty.batch.core.BatchRenderBackendCoreProfileInternal.addImageToAtlas(BatchRenderBackendCoreProfileInternal.java:218)
    at de.lessvoid.nifty.renderer.lwjgl.render.batch.core.LwjglBatchRenderBackendCoreProfile.addImageToAtlas(LwjglBatchRenderBackendCoreProfile.java:106)
    at de.lessvoid.nifty.batch.BatchRenderImage.uploadImageToAtlas(BatchRenderImage.java:197)
    at de.lessvoid.nifty.batch.BatchRenderImage.upload(BatchRenderImage.java:142)
    at de.lessvoid.nifty.render.NiftyImageManagerExtBatch$ReferencedCountedImageBatch.upload(NiftyImageManagerExtBatch.java:169)
    at de.lessvoid.nifty.render.NiftyImageManagerExtBatch.uploadScreenImages(NiftyImageManagerExtBatch.java:69)
    at de.lessvoid.nifty.render.NiftyImageManager.uploadScreenImages(NiftyImageManager.java:60)
    at de.lessvoid.nifty.render.NiftyRenderEngineImpl.screenStarted(NiftyRenderEngineImpl.java:859)
    at de.lessvoid.nifty.screen.Screen.startScreen(Screen.java:216)
    at de.lessvoid.nifty.Nifty.gotoScreenInternal(Nifty.java:714)
    at de.lessvoid.nifty.Nifty.gotoScreen(Nifty.java:665)
    at de.lessvoid.nifty.examples.defaultcontrols.ControlsDemo.main(ControlsDemo.java:89)`

any idea?

ghost:
Thanks for merging. Yeah, I didn't know about the CoreGL library, funny that it has the same name as the interface I created. Regarding the bug, I'm on it. It would be helpful to know which texture id is being passed to the bindAtlasTexture there, since what's happening is it's saying that texture id doesn't exist / can't be found.

ghost:
@void256 I could submit a patch with some debugging code added in, then all you would have to do is run it and let me know what it says. Might be easier than you having to do it yourself.

ghost:
@void256 Also, if you could please enable fine debugging and let me know the output.

void256:
@3xp0n3nt send you the log by mail - it was too large for pastebin :D

Issue #218, opened 2014-02-28
setConstraint to default won't have any effects


relu91:
If you set a constraint with SizeValue.def() the layout doesn't take care about it. For example in absolute layout if you setConstraintY(SizeValue.px(200)) and then setConstraintY(SizeValue.def()) the excepted behavior should be to see the element in y=0 location but it isn't even if you call layoutElements() in the parent of element. 

the problem could be in these line on absolutelayoutManager : 

 if (cons.getY().hasValue()) {
        box.setY(rootBoxY + cons.getY().getValueAsInt(rootBoxHeight));
  }

Maybe I'll try to change these :+1:

void256:
looks good! thanks for the fix too :smile:

PR #219, opened 2014-03-07
Solve issue #218 

relu91:
This are just some little changes but they make attributes removal little easier to do in nifty-editor I haven't checked the vertical behavior yet , but the others seems fine.


relu91:
Thanks! :) i've cut&pasted the comment and didn't notice that... sorry!

Issue #176, opened 2013-12-23
TabGroup should have a method to retrieve the tab at a specific index


Entropy4711:
Something like:

Tab getTabAtIndex(int index)

Issue #175, opened 2013-12-23
TabSelectedEvent has only private getters


Entropy4711:
If you want to know which Tab has been selected in a TabGroup via the TabSelectedEvent, you cannot access the Tab or the tab-index from the TabSelectedEvent because all getters are private. This makes this Event useless.

Issue #53, opened 2012-09-24
RFE: Support for Double-Click/Pointing at


mkaring:
Currently the Nifty-GUI only supports single mouse clicks, but it lacks the support for double clicks. So in case the user clicks twice on a location the single click function should **not** be executed, but the one for double clicks should be executed.

Furthermore it would be handy to have a pointing at functionality that is triggered one the user stops the mouse cursor on top of a element. This does only trigger in case the mouse cursor stops on the element, unlike the hover function that triggers already in case the mouse cursor moves over the new element.

rhavin:
Usually, on a doubleclick the singleclick-fn HAS to get executed first! In Windows – for example – if you douleclick, the first click selects the item (like a singleclick) and the second click executes it. Having items react to a singleclick other that to the first click of a doubleclick breaks users assumptions about GUIs. It would also make the GUI respond significantly slower, because it had to wait for the defined time of a doubleclick before it could execute any singleclick-event. I you really want your application to behave completely different on a singleclick than on the first click of a doubleclick – i'd strongly suggest not to do that – you could trigger a timer on first click that executes your code after a given time and have the doubleclick-handler remove the timer before it executes. But i state it again: it breaks users assumptions about GUIs.
Of course, modern mouses got a lot of keys (i've got 6) and there are also the modificator keys Ctrl and Shift, so if you really need that much of different behaviour, why not use them?

philipwhiuk:
> Furthermore it would be handy to have a pointing at functionality that is triggered one the user stops the mouse cursor on top of a element. This does only trigger in case the mouse cursor stops on the element, unlike the hover function that triggers already in case the mouse cursor moves over the new element.

This certainly isn't in the AWT events. There would have be some configurable timeout to determine how long a stop is. The normal behaviour is `mouseover`/`mouseout` (HTML), `mouseEnter`/`mouseExit` (AWT), `MouseEnter`/`MouseLeave` (WinForms)

void256:
relu91 added support for this now, see the commits referenced in the issue. Currently there is still a little issue with first click, see this commit for details: 9acfdf2da377ceb0b7a2bbc89467092bfb994c2d

I'll close this issue for now anyway!

relu91:
[...]

PR #220, opened 2014-03-27
Some issues fixed

relu91:
Hi , I worked on this issues #53 , #176 and #175 . About #53 this is a first implementation indeed it works only for primary button but I tested it and it worked fine . The behavior is : click-click-click will fire onPrimaryClickEvent - MultiClickEvent( count = 2 ) - MultiClickEvent( count = 3 ) . I used this implementation because is similar to swing one . I leave "single click event" because I don't know if this will in some way break the current users assumptions . 
I see in the commits that two entire classes are overridden I don't know why .... do you know how I could solve this ?

### DIFF
--- a/nifty-controls/src/main/java/de/lessvoid/nifty/controls/TabGroup.java
+++ b/nifty-controls/src/main/java/de/lessvoid/nifty/controls/TabGroup.java
@@ -174,4 +174,13 @@ public interface TabGroup extends NiftyControl {
    * @throws IllegalArgumentException in case the tab that is supposed to be removed is not a part of this tab group
    */
   void setTabCaption(@Nonnull Tab tab, @Nonnull String caption);
+  
+    /**
+   * Get the tab control at a specified index.
+   *
+   * @param index the index of the tab control
+   * @return the tab control
+   */
+  @Nullable
+  Tab getTabAtIndex(final int index);
 }
--- a/nifty-controls/src/main/java/de/lessvoid/nifty/controls/TabSelectedEvent.java
+++ b/nifty-controls/src/main/java/de/lessvoid/nifty/controls/TabSelectedEvent.java
@@ -51,7 +51,7 @@ public class TabSelectedEvent implements NiftyEvent {
    * @return the tab group
    */
   @Nonnull
-  private TabGroup getParentGroup() {
+  public TabGroup getParentGroup() {
     return group;
   }
 
@@ -61,7 +61,7 @@ public class TabSelectedEvent implements NiftyEvent {
    * @return the tab that is now visible or {@code null} in case no tab is visible
    */
   @Nonnull
-  private Tab getTab() {
+  public Tab getTab() {
     return tab;
   }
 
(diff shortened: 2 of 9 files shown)


void256:
Good work! I'm ok with commit 92727ce :smile:

But I can't apply 9acfdf2 as-is because I can't see what you have changed, I'm sorry. The ElementInteractionClickHandler, MouseClickMethods and PrimaryClickMouseMethods classes look like as you would have changed each line. Can you please check this again? It might be just a "new line" setting in your IDE or did you reformat the complete file eventually? Not a good idea because one can't see what has changed in these files at all :smile:

And I think it would be helpful if you could merge your branch before applying your changes and creating the pull request. So that only your changes appear in the pull request?

relu91:
Ok @void256 , sorry about https://github.com/void256/nifty-gui/commit/9acfdf2 , I think is eclipse fault , because I only edited some lines of those files. So do you know which are the steps to revert my changes safely ? I mean not completely re-write all the commit but only edit those files . About merge , I used github client , I opened a console e simple fetch your repo and then merge mine with yours... but I see that now, in the pull there's also your commits do you know why? where I am wrong ? .
Sorry about this noob questions but every time I feel that I know everything about git , suddenly I found that I am completely wrong! :D 

Thanks

relu91:
Oh , i forget an other question :P . Is there a way to store some global values in Nifty-gui? . I know that it's hard to see but in this commit https://github.com/void256/nifty-gui/commit/9acfdf2 I used a threshold value to know when I stop to record multi click events . I think that this value should be configurable by the nifty user . Any suggestion ? :)

void256:
**About global values**
There is nifty.getGlobalProperties() which returns a Properties instance which you could use for that purpose :smile: It was not really meant for that because it's purpose was to store key/values for special values replace mechanism using the `${PROP.key}` in attributes BUT I don't really see why you can't use it for your purpose as well :smile:

**About fixing your pull request**
I'm not sure if you can do it without rewriting your changes. But I think you should be able to pull the original branch into yours which should make my commits disappear from your pull request, something like: `git pull nifty-origin 1.4` I'm not sure how you do that and especially how your remote is called (nifty-origin). But the basic idea is to pull in all of the 1.4 changes into your fork before creating the pull request, I think.

About fixing the new lines in these files ... I'm not sure how you can do that without rewriting your changes. Can you diff them local and see what you've changed? In that case I would suggest reverting your changes and then reapplying them manually without changing the new lines automatically. I don't think there is an automatic way to do that tho...

relu91:
Thanks! this afternoon I'm going to solve this. I also found out why those three file are all modified ! I used auto format code in eclipse.... and it might change the whole file ... shame on me -.-" :)

relu91:
It seems that git hates me ... On my local repo all is fine ... but when i push in my remote forked repository the whole file is changed -.-" . Maybe should I  close this pull request first ?

NetzwergX:
Looks like an issue with line endings? https://help.github.com/articles/dealing-with-line-endings

relu91:
I don't if it was line endings , I changed an option in my git client ( Use rebase for pull ) . Now files are ok :) . Is it ok even if there are a couple of new commits ? anyway thanks @NetzwergX

void256:
Looks better now @relu91 :smile: Thanks for cleaning it up!

PR #222, opened 2014-04-01
Finish multiclick support

relu91:
This commit adds multiclick support for all nifty mouse buttons. This time with 90% of mess in less ! :+1:


void256:
Yep, looks a lot better without the reformating of whole source files :+1: :smile:

Issue #138, opened 2013-08-19
[1.4] childClip inside childClip does not work correctly


void256:
Thinking about [this thread](http://hub.jmonkeyengine.org/forum/topic/nifty-gui-scrollpanel-problem/) at the jme forum I've found the following code:

``` java
public void enableClip(final int x0, final int y0, final int x1, final int y1) {
    // in case there already is a clipping area set we can't override it. not sure if this is correct tho but it
    // fixes issues when some parent element has set a clipping area in which child elements should not be able to override the area.
    if (clipEnabled) {
      return;
    }
    updateClip(true, x0 + getX(), y0 + getY(), x1 + getX(), y1 + getY());
  }
```

I think the right way to solve that is to clip the new clip rectangle to the existing one so that we can make sure nothing is drawn outside the existing clipping rectangle and still be able to further restrict the clipping. Please note that it's still not possible to override an existing clipping rectangle. But thats ok since that's what this comment tries to do anyway.

PS: We can get the code to intersect rectangles from the batched renderer where we already solved that.

relu91:
I don't know if I should open a new issue but I comment here first. The implementation that was done prevent final user do dynamically change the clipArea of nifty. I don't know if this would be a common user case , but it is mine :-) . For example if I simply want to move the clipped area around after a increment of a scrollbar , the clip becomes littler and littler instead of move at desired position. My suggestion is to add a method to force the clip instead of using enableClip , you could use setClip to force a clip without cropping it . This is also Java2d approach see : http://docs.oracle.com/javase/7/docs/api/java/awt/Graphics2D.html#clip(java.awt.Shape) and http://docs.oracle.com/javase/7/docs/api/java/awt/Graphics.html#setClip(java.awt.Shape) .

void256:
Yes, that would be possible :smile: We would need a way to enable something like "absoluteClip" at the element and find a way to route that information down until the NiftyRenderDevice. Once at NiftyRenderDevice level it's easy because when that "absoluteClip" information is present all of the default behaviour would be overwritten and the new clip rectangle is used as-is :D

I would suggest not using the Java2d API directly but find something similar that fits better into the existing "clipChild" attribute - because I would left the default behaviour as-is. Maybe introduce a separate attribute for that, like "clipChildAbsolute" or so?

relu91:
[...]

PR #223, opened 2014-04-15
Absolute clipping

relu91:
I hope that this pull request makes things clearer ! you could always ask me further changes! I've also added a simple working example :) hope it helps #138

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/Nifty.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/Nifty.java
@@ -349,7 +349,7 @@ public class Nifty {
     if (clearScreen) {
       renderEngine.clear();
     }
-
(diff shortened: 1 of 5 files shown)


void256:
This looks good! Thanks! I'll merge it in a second :smile:

The only minor drawback I see is that now we'll run with clipping **always** enabled - even in cases where no one has enabled absolute clipping! Depending on the specific RenderDevice implementation this may or may not mean a little performance cost. I really think it's not much so it's okay but in general it would be better if we would only enable clipping when someone has set the absolute clipping. Just in case you'd like to take a look at this again.

relu91:
Cool! 
I though about that , but I was waiting for your feedback :smile: and I didn't implement it . I'm going to add it as soon as I finish my work on 0.5.9 version of nifty-editor :smile: .

PR #233, opened 2014-05-22
Add subImage-Resize for ImageMode , fixes #225

### LINKED ISSUE #225: Add Image mode subImage-resize
I did UI design recently, I found it was really annoying to split my UI image to small image to use resize mode in Nifty GUI.
Now we only have sprite-resize mode, but it is actually useless when we need components in different size.

relu91:
This works like sprite-resize mode :
imageMode="subImage-resize:x,y,w,h,x0,x1,x2,y0,x3,x4,x5,y1,x6,x7,x8,y2"


void256:
@relu91 EXCELLENT work! :smile: You even added an Issue225 example! Very well done! :thumbsup:

PS: This commit is so excellent that I won't complain about the unrelated organize imports in Clipping.java ... :wink:

PR #235, opened 2014-05-23
New styles with sounds and costum effects, fixes #228

### LINKED ISSUE #228: <registerSound> not works in style and control xml file
If I have many sound effect defined in different style xml files, I have to register these sound files in my main screen xml file
I think it is not a good design rule. It's better to register the files when I need them.

relu91:
I don't like study physic so I'm sending pull requests :smile: . This will lead to a very rich styles since a final user can also embed sounds and effects in a style definition . There's no builder support at the moment but I don't think that is an actual user case since he could register sounds and effects using the nifty instance . Also I think that an user will be able to pack : styles , sounds and effects class in a single jar and then redistribute it , I didn't test it though . Finally I built also an example , try it :dancer:


void256:
wow! looks nice as well! great job again! :thumbsup:

PR #236, opened 2014-06-03
Example of issue 100

relu91:
A simple main to test issue 100 . You can choose between Italian , German , French and the system default.

### DIFF
--- a/nifty-examples/nbactions.xml
+++ b/nifty-examples/nbactions.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<actions>
+        <action>
+            <actionName>run</actionName>
+            <goals>
+                <goal>process-classes</goal>
+                <goal>org.codehaus.mojo:exec-maven-plugin:1.2:exec</goal>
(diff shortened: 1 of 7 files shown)


void256:
I've found my stash and it already had an example for issue 100. I'll close this for now and hope you don't mind :smile:

void256:
... but not all is lost! I've used the DropDown from your example now since mine only had a couple of buttons :smile:

relu91:
Ok ok ! :smile: I didn't spend so much time on this, so don't worry ! I hope that you can fix the issue . Also would be cool that even that the other "special values" can be automatically updated when you change them on the java side . Like when you change a property in globalProperties also all the elements receive that change , but I think this is even more difficult to achieve since it requires a sort of DataBinding ...  Well just some thoughts :smile:

void256:
Yes, would be nice but I don't see that happening for all attributes - it's getting too complicated :/

I have it working for TextRenderer now which means all texts can be replaced when the locale is being changed :smile: I'm about to commit that in a couple of minutes ...

Issue #133, opened 2013-07-16
RFE: OnClick should trigger at button release


mkaring:
Currently the mouse onClick function is triggered when the mouse button is pressed down. This clashes with most other GUI frameworks. Usually onClick is fired when the mouse button is released.

It should be changed in general or at the very least there should be a option to allow to change this.

void256:
I agree and will add that. I already have a plan for it :wink:

ghost:
#256 is indirectly related to this issue.

PR #256, opened 2014-07-30
Fix NiftyStandardInputEvent.Activate bug.

ghost:
NiftyStandardInputEvent.Activate only simulates a mouse click, but not a
mouse release. This causes a bug where only listening for
ButtonReleasedEvent's but not ButtonClickedEvent's for a specific button
will cause input mappings to NiftyStandardInputEvent.Activate to not
work with that button. This results in undesirable & unexpected behavior
such as pressing the enter key on a currently focused button does not do
anything.

The solution is to make NiftyStandardInputEvent.Activate simulate a
mouse click and a mouse release, so that when
NiftyStandardInputEvent.Activate is triggered, a ButtonReleasedEvent is
fired in addition to the ButtonClickedEvent.


ghost:
This is indirectly related to #133, which is the real solution to this bug, which is caused by separating the idea of a "click" from a "release".

PR #257, opened 2014-07-30
Allow SizeValue floating point values.

### LINKED ISSUE #244: SizeValue throwing exceptions for floating point percentage values
If I interpret this\* snippet of the 1.4 changelog correctly, it says that some elements round SizeValues to get ints, but floats are still used internally. Which makes it a bit paradoxical that SizeValues now throw Exceptions when I attempt to make them take a float percentage, and since I rely on the floating parts to position many elements exactly, this is also quite annoying to me..
- Quote:
  Many classes in the effects and layout packages in nifty-core
  are using a method in the SizeValue class called getValue that returns a
  floating point value. The issue is not with this method itself per se,
  but with callers of the method, who really need an integer value, and
  cast the returned floating point value to an integer, effectively
  discarding everything to the right of the decimal. It can easily be
  verified that calling getValue and then casting to an integer returns
  incorrect values much of the time. This bug is now verified with a
  simple test in SizeValueTest. The bug rears its ugly head the worst when
  converting a simple pixel value of say "212px" from a SizeValue whose
  percent is set to 100, to a floating point pixel value, and then casting
  it to an integer: the result will always be 211 and NOT 212. There are
  countless other values that return a similar result when the SizeValue
  percent is set to 100.
- The solution to this error, which rectifies issues such as creating
  quads at off-by-one pixel sizes, is to change all callers of getValue
  that actually need an integer value to use the preexisting getValueAsInt
  method, BUT since it ALSO simply casts the floating point value of the
  getValue method to an integer, persisting the bug, we modify the
  getValueAsInt method to use Math.round, and round the result returned by
  getValue to the nearest integer instead of casting it. This returns much
  more correct values, and fixes the bug. This solution has been
  integration-tested without noticing the introduction of any other
  regressions.

Just for completion, here's an xml snippet that used to work perfectly in 1.3.2:
        <panel x="65%" y="1.5%" width="17.4%" height="5.67%" childLayout="center" >
          <interact onPrimaryClick="scrollUp()" />
          <text text="${dialog.configureKeys}" color="#000f" style="baby-base-font" />
        </panel>

```
    <panel x="15%" y="78.5%" width="20%" height="9%" childLayout="center" >
      <interact onPrimaryClick="close()" />
      <text text="${dialog.close}" color="#000f" style="baby-base-font" />
    </panel>
    <panel x="69%" y="78.5%" width="20%" height="9%" childLayout="center" >
      <interact onPrimaryClick="acceptChanges()" />
      <text text="${dialog.acceptChanges}" color="#000f" style="baby-base-font" />
    </panel>


    <panel x="65%" y="-6.67%" width="17.4%" height="5.67%" childLayout="center" >
      <interact onPrimaryClick="scrollDown()" />
      <text text="${dialog.settings}" color="#000f" style="baby-base-font" />
    </panel>
```

ghost:
Modify the internal storage of SizeValue to use a float rather than an
int to allow for floating point values in SizeValue strings.

Add SizeValue tests to verify that floating point strings are accepted
as valid values and are parsed correctly.

SizeValue.getValueAsInt now rounds the floating point pixel values to
the nearest integer using Math.round, which is the safest way to do it.
It is also consistent with the already existing behavior for floating
point percentage values, which already uses Math.round in the
getValueAsInt method (it was added to fix a subtle truncation bug
causing off-by-one pixel errors).

Fix failing tests in
nifty-core/src/main/test/java/de/lessvoid/nifty/layout package by
verifying resulting SizeValue strings as a decimal value instead of an
integer value. For example, new SizeValue("100px").toString() will now
equal "100.0px" instead of "100px".

The SizeValue.toString method is the only place where the extra ".0" in
(for example) "100.0px" as a string will come into play because even
though most of the time it will be a whole number, now there is the
chance that it could be something like "100.5px". It seems to be the
most reasonable behavior to always have the decimal and trailing zero
for "whole number" SizeValue.toString outputs, rather than only having
the decimal when it is not ".0". This could be considered a significant
change, but it doesn't break anything because only the output format is
changing - SizeValue still accepts "100px" (with no decimal) as a valid
value, it just converts it to "100.0px".

The SizeValue equals & hashcode methods were deleted and replaced by
auto-generated versions from IntelliJ, which are safe because they both
rely on the Float.floatToIntBits under the hood for the internally
stored float value. Tests have been added to SizeValueTest to verify
that equals is working correctly. For example, new SizeValue ("10.0px")
should be equal to new SizeValue ("10px"), but new SizeValue
("10.0001px") should not be equal to new SizeValue ("10px") nor new
SizeValue ("10.0002px").

Fixes #244.

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/tools/SizeValue.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/tools/SizeValue.java
@@ -61,7 +61,7 @@ public final class SizeValue {
    * The value stored in this size value, check {@link #hasCalculatedValue} and {@link #hasValue} to ensure that this
    * value is valid.
    */
-  private final int value;
+  private final float value;
 
   /**
    * This is set {@code true} in case the {@link #value} is set as calculated value.
@@ -86,7 +86,7 @@ public final class SizeValue {
       throw new IllegalArgumentException("Size value type " + type.name() + " requires a value!");
     }
     this.type = type;
-    value = 0;
+    value = 0.0f;
     hasValue = false;
     hasCalculatedValue = false;
   }
@@ -151,17 +151,17 @@ public final class SizeValue {
     hasCalculatedValue = false;
     if (valueParam == null || valueParam.isEmpty() || valueParam.equals("default")) { // alias for "d"
       type = SizeValueType.Default;
-      value = 0;
+      value = 0.0f;
       hasValue = false;
       return;
     } else if (valueParam.equals("sum")) { // alias for "s"
       type = SizeValueType.Sum;
-      value = 0;
+      value = 0.0f;
       hasValue = false;
       return;
     } else if (valueParam.equals("max")) { // alias for "m"
       type = SizeValueType.Maximum;
-      value = 0;
+      value = 0.0f;
       hasValue = false;
       return;
     }
@@ -178,7 +178,7 @@ public final class SizeValue {
       // no suffix -> falling back to px
       type = SizeValueType.Pixel;
       try {
-        value = Integer.valueOf(valueParam);
+        value = Float.valueOf(valueParam);
         hasValue = true;
       } catch (NumberFormatException e) {
         throw new IllegalArgumentException("String value [" + valueParam + "] does not fit the required format.", e);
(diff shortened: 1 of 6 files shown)


void256:
Great work again! :thumbsup: :smile:

PR #265, opened 2014-08-06
Catching a possible NPE on linux systems

mkaring:
This problem may be caused by a bug in the java runtime.

For further information: JDK-6322854


void256:
wow - I guess this really happend? :wink:

Thanks for fixing it!

PR #302, opened 2014-12-27
Renamed "CSS" to "Style" as discussed with void256

spaceemotion:
From the commit message:

> This renames the CSS related part of the system to be known as "styles". Since its not a proper CSS file as a browser would use it, it's a nifty-specific CSS system.
> 
> The original files regarding default CSS handlers and systems have been left untouched (like the parsers and external libraries).
> 
> Javadocs have been changed accordingly as well.

_Discussion in context: http://void256.github.io/nifty-gui/blog/update/nifty-2.0-css/#comment-1760677275_


nifty-examples/src/main/resources/logging.properties
```diff
@@ -6,8 +6,8 @@ java.util.logging.ConsoleHandler.formatter = de.lessvoid.nifty.examples.usecase.
 # enable fine logging of input device
 #de.lessvoid.nifty.lwjgl.NiftyInputDeviceLwjgl.level = ALL
 
-# Nifty CSS processing
-de.lessvoid.nifty.internal.css.NiftyCssAnnotationProcessor.level = FINE
+# Nifty style processing
+#de.lessvoid.nifty.internal.style.NiftyStyleAnnotationProcessor.level = FINE
```
spaceemotion, line 10: This class did not exist on my end so I commented out this one.

spaceemotion:
While working with the source code IntelliJ pointed out a few things than can be changed to gain performance (namely the tons of for loops that are still using and index instead of an iterator). Should I make a separate PR for that or would you like to address those issues by yourself?

void256:
I would argue that IntelliJ is wrong in that case. I remember changing all Iterator based loops in Nifty 1.x to get rid of the AbstractIterator instance that this kind of for loop generates. In Nifty 1.x this was a major reason that Nifty allocated tons of new instances each frame which in turn created a lot of GC runs in the end.

After changing that I usually prefer the simpler index based access which doesn't create unnecessary garbage. One could argue that this is an optimization that is too early and I'd probably agree with that too :smile: but I don't think I'd want to change these loops now back to iterator based for loops!

spaceemotion:
Well as long as you're using Array Lists and that kind of stuff, it won't be too much of a problem, but once you use things like Linked Lists getting elements by their index can slow things down.

NetzwergX:
There are several Android versions who leak the iterator in for loops. depending on the backing-implementation, index-based iteration is as fast or faster then iterator-based iteration. Especially when doing games in java you have to be cautious about object count. the more you allocate, the longer the gc cycle, the lower the fps. New JVMs tend to be better - you probably get lucky and the iterator is allocated on the stack due to escape analysis - but older JVMs will have to clean it up.

void256:
Thanks @NetzwergX and thanks @spaceemotion for the pull request too :+1:

spaceemotion:
Huh, didn't know that thing about android and JVMs, thanks for the tip @NetzwergX ;)

NetzwergX:
I only knew about android leaking iterator due to JME beign heavily hit by that. They had to refactor  lots (and I really mean lots) of code to get rid of iterator-based loops to avoid crumbling under android. I don't know if it is still an issue under more recent android versions since I have not followed up on that - so depending on what versions you want to support/target you might not be hit by this.

PR #313, opened 2015-01-31
Update SinusPulsator to accept "reverse" parameter

Davi-DeGanne:
The "reverse" parameter allows the effect to begin in the applied state, progressing towards the unapplied state. Useful for ImageOverlayPulsate effect to fade out from an ImageOverlay effect.


void256:
Looks good to me! Thanks :smile:

PR #319, opened 2015-03-12
Adding nifty-renderer-processing to nifty

barefists:
Implementation of RenderDevice and InputSystem for Processing 2.*

A simple implementation of RenderDevice for Processing wrapped around the Java2D renderer, but using Processing's own VLW bitmap font format. Also included is an implementation of InputSystem for Processing's proprietary input event classes.

I also needed to change access of AwtToNiftyKeyCodeConverter in the Java2D renderer package to public for use in implementing the input system. Let me know if this is okay.


spaceemotion:
Are the eclipse files necessary?

barefists:
Oh, not really. I'll add them to .gitignore - sorry about that.

bgroenks96:
@void256 Why has this PR been sitting here for so long?

void256:
Waaaah! I think I agreed with spaceemotion that the eclipse files shouldn't be part of the commit .. and then somehow didn't notice barefists changes!

My apologies!

dominic-jones:
I am still noticing eclipse files in the nifty-minim-soundsystem directory. Also, mvn clean install is not running from the project root. Claims it cannot find that new soundsystem module. Might be my setup though. Investigating.

void256:
_argh_ :smile: already cleaned that up too!

barefists:
Seems processing is gradually depreciating Java2D over JavaFX in the latest 3.0.1 version. Guess it's time to revise the Processing renderer to better support the changes. If you are trying to use the current renderer in 3.0.1, you'll run into issues, it still works well for 2.2.1 though.

PR #339, opened 2015-05-22
log debug messages as fine instead of warn

### LINKED ISSUE #337: TODO Remove this debugging statement
First of all, I just wanted to say I really like nifty-gui and really appreciate all the work you've saved me. =)

Some of the log messages were confusing to me though. I kept getting warnings about textures, but I couldn't figure out what I was doing wrong. After glancing at the source, it looks like there are a few warnings that aren't really warnings at all in the [batch rendering system](https://github.com/void256/nifty-gui/blob/1.4/nifty-core/src/main/java/de/lessvoid/nifty/render/batch/core/BatchRenderBackendCoreProfileInternal.java#L388). It looks like these warnings might never have been intended to be released at all. They're all marked with comments like: "TODO Remove this debugging statement."

Should these warnings be removed? I can send a PR to remove them if that would make it easier for you.

cuchaz:
there's no issue an end-user needs to resolve, so the warnings are confusing

fixes #337


bgroenks96:
Looks great, thanks! :+1:

Issue #341, opened 2015-05-28
Hint effect does not set visibleToMouse; Hint layer not deactivating


dominic-jones:
On upgrading to 1.4.1 that Hint hover text had ceased displaying a hover layer. It appeared as if the UI element had no hint effect at all.

Some searching led me to http://sourceforge.net/p/nifty-gui/bugs/53/, and on the surface this is a regression of that.

In testing similarity with that bug, I tried enabling visibleToMouseEvents. The hint layer activates correctly but then fails to deactivate.

dominic-jones:
I have investigated this locally and have a fix I will submit as a pull request.

void256:
I've applied the merge request and guess it's ok to close this issue now! Please re-open if I've missed something.

PR #342, opened 2015-05-28
fix: #341 Correcting Hint Effect layer not activating / deactivating

dominic-jones:
In Effect, the hoverEffect boolean was being checked as a prerequisite to enabling visibleToMouseEvents. During the refactor in bd08f8ce208a2b8e626d1e5f467beb016ac93360 this check was moved. It now happens before anything that can set hoverEffect to true, and in effect evaluates to always false.

This change drops the if check entirely and moves enabling the visibileToMouseEvents flag to Effect#enableHover(). Not sure if that's where you want it, but it seems reasonable.
# 

This also corrects a check in the Hint effect itself. When the Hint effect deactivates, it checks if there is exactly one child, and if so, hides the hint panel. I do not know the cause of this one, but there were zero children at this check. (I assume some other code changed that cleaned up the children).

This was changed to check if there is one or less children as a condition to hiding the hint layer.
# 

Normally I'd like to put these fixes up separately, but since I went in to fix the hint layer because of the first bug, and then discovered the second one, I'd rather commit it together and put the hint layer back into a working state.
# 

It is worth noting that if a hover effect is removed from the element, I think the visibleToMouseEvents flag will remain set. That might be a separate issue where we apply a check to see if visibleToMouseEvents was set exclusively because of adding a hover effect. Uncertain if this is required though.

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/effects/Effect.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/effects/Effect.java
@@ -95,15 +95,13 @@ public class Effect {
     controllerArray = controllers.toArray();
     effectEvents.init(nifty, controllerArray, parameter);
     customFlag = false;
-
-    if (hoverEffect) {
-      element.setVisibleToMouseEvents(true);
-    }
   }
 
   public void enableHover(final Falloff falloffParameter) {
     hoverEffect = true;
     falloff = falloffParameter;
+
+    element.setVisibleToMouseEvents(true);
   }
 
   public void disableHover() {
--- a/nifty-core/src/main/java/de/lessvoid/nifty/effects/impl/Hint.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/effects/impl/Hint.java
@@ -152,7 +152,7 @@ public class Hint implements EffectImpl {
       hintPanel.markForRemoval(new EndNotify() {
         @Override
         public void perform() {
-          if (hintLayer.getChildrenCount() == 1) {
+          if (hintLayer.getChildrenCount() <= 1) {
             hintLayer.hide();
           }
         }
--- a/nifty-core/src/test/java/de/lessvoid/nifty/effects/HoverEffectTest.java
+++ b/nifty-core/src/test/java/de/lessvoid/nifty/effects/HoverEffectTest.java
@@ -1,0 +1,57 @@
+package de.lessvoid.nifty.effects;
+
+import de.lessvoid.nifty.Nifty;
+import de.lessvoid.nifty.effects.impl.Nop;
+import de.lessvoid.nifty.elements.Element;
+import de.lessvoid.nifty.spi.time.impl.AccurateTimeProvider;
+import org.easymock.EasyMock;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.LinkedList;
+import java.util.Properties;
+
+import static org.easymock.EasyMock.expectLastCall;
+import static org.easymock.EasyMock.replay;
+
+public class HoverEffectTest {
+
+  private static final boolean INHERIT_FALSE = false;
+  private static final boolean POST_FALSE = false;
+  private static final boolean OVERLAY_TRUE = true;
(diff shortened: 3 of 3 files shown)


dominic-jones:
Please give me some feedback if there are concerns or changes around the content of this PR.

bgroenks96:
Everything looks good, your branch builds fine with a fresh 'mvn clean install'

I'm having issues running tests at the moment... weird classpath errors I'm not willing to solve right now.

Were you able to run any tests and verify the hover effect is working?

If you can, try testing it in a couple of the big demos (AllExamples, AllControls, etc).

dominic-jones:
I have tried running the samples in general and they are fine. (To a given point: out of the box, the libgdx sample works for me. With or without this change, neither the JOGL or Slick demos work for me as-is).

However, I don't think these demos are helpful here. I don't know them (or the Nifty code) at all well. From what I can tell with the right searches, the only place a Hint is defined for use is in the MenuButtonControlDefinition class. I can see this class registered by the ControlsDemo, but I cannot find a reference to it in any Nifty screen xmls (and it seems there are no ScreenBuilder examples at all).

So... yes the demos work, but there are no demos for hints (to the best of my knowledge). I know it works (now) in my own project (because that's how I found the break on upgrading from 1.3.3 in the first place).

void256:
The ControlsDemo does use the Hint effect to display a Hint for the top menu buttons. I've checked your commit with that example and it works for me.

I'll apply this merge request now. It looks valid to me. Thanks for figuring it out! :smile: 

I think what Martin tried to do with his original change was that any element you apply a hover effect to should automatically be visible to mouse events. Otherwise the hover effect would not work at all unless you manually set visibleToMouse to true too.

But I agree that this didn't work as expected and your change actually makes it work again.

I'm not sure about the checking of the hintLayer children count though but I think it does no harm so I'm ok with that as well.

It would have been nice if you could have added a SSCCE that reproduces the issue and shows that the fix actually fixes the issue but besides that it looks ok to me!

dominic-jones:
His change worked great actually, but a later refactor changed the ordering of the code which prevented the layer activating.

As for the hintLayer children count, I do not know that area well at all. My best assumption there is that if it only has one hint on disposal it thinks the layer can do. But when I was stopping there with a breakpoint it was routinely 0. I imagine the hint itself was being cleaned up elsewhere (perhaps as a result of another refactor).

What manner of SSCCE would you have liked here? An extra example showing off just a hint working? A gist / link / repo of an SSCCE outside the project? I know how I'd build one, just not sure how you'd want to receive it. This may not be my only contribution, so I'd like to learn and work within your patterns in future.

void256:
In the best case a SSCCE would be so small that it could fit into the pull request comment. I mean here on github not the actuall commit comment of course :wink:

It could depend on the NiftyExampleLoaderLWJGL or so. That would be okay. Just anything that would make it simple to copy and paste it somewhere and execute it. Just a dedicated example Main class or ScreenController that reproduces the issue in question. In your case yes, a hint example showing the hint not working correctly without your fix.

That being said, we've been discussing a dedicated nifty-sscce repository (or similar) for such things.

I have locally created seperate Maven projects for some of the issues but they are currently not commited/pushed somewhere. Having a dedicated repo would allow users to contribute example code with resource files too if necessary. But right now we don't have that yet but I still think it's a good idea!

@bgroenks96 Do we want a dedicated project on github for that? Should I create one? What do you say? I could add the issues I already have projects for as examples too ...

dominic-jones:
I've generally written my own repos, or clones to do it. However a dedicated repository with some minimal setup might reduce the barrier to entry for good examples. It's all down to how much you want contributors to maintain :)

bgroenks96:
@void256 For something like nifty-gui where there is so much room for weird little corner case bugs that require SSCEs to repro, yeah that sounds like a pretty good idea to me.

Could also do it as a git sub-module if you're a masochist :laughing:

void256:
New repo added here: https://github.com/void256/nifty-gui-issues

And I've added the two issues I had on my local machine as an example.

bgroenks96:
Cool. I'm sure that will be useful.

PR #344, opened 2015-05-29
PR#319 Cleanup

dominic-jones:
Cleans up code from https://github.com/void256/nifty-gui/pull/319

As checked in, mvn clean install fails as the module referenced in the parent pom is incorrect.
- Removes Eclipse files
- Corrects the name of the soundsystem referenced in the parent pom


dominic-jones:
Oh, you already beat me to it :)

void256:
Hehe ... but only by a couple of seconds :wink: Thanks :smile:

bgroenks96:
Werid. Mvn clean install didn't fail for me when I ran it. Which module did it fail on?

EDIT: Oh... I just realized #319 was that other dinosaur PR.

PR #352, opened 2015-07-22
feat: system property setting for default NiftyInputMapping.

### LINKED ISSUE #348: Disable "tab key" behavior?
Nifty uses the tab key to traverse through elements and set their focus. Any way to disable this? I need to use the tab key for my game itself.

bgroenks96:
<<< PLEASE TEST THIS! >>>

ElementType will now use a default class type accessed by name via a system property specified by NiftyDefaults.

Unfortunately, there is no way to inject the setting into ElementType at this time due to no instance of Nifty (or Screen, ScreenController, etc) being accessible. This is a "maximum compatibility" approach, but it should be noted that it is most certainly non-optimal. Using system properties is kind of gross...

Regardless, this should allow users to override Nifty's default input mapper, so that mappings can be applied globally.

Note: the danger here is that the setting will persist throughout the JVM rather than just a single Nifty or Screen instance. Again, not optimal.

Resolves #348

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/Nifty.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/Nifty.java
@@ -36,6 +36,7 @@ import de.lessvoid.nifty.elements.Element;
 import de.lessvoid.nifty.elements.ElementMoveAction;
 import de.lessvoid.nifty.elements.ElementRemoveAction;
 import de.lessvoid.nifty.elements.EndOfFrameElementAction;
+import de.lessvoid.nifty.input.NiftyInputMapping;
 import de.lessvoid.nifty.input.NiftyMouseInputEvent;
 import de.lessvoid.nifty.input.keyboard.KeyboardInputEvent;
 import de.lessvoid.nifty.input.mouse.MouseInputEventProcessor;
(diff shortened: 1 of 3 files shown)


bgroenks96:
@void256 Can I get some feedback on this?

void256:
Sorry @bgroenks96 - forgot to comment :grin:

Thanks for the pull request! This already looks good!

I'd vote to drop the system property though. What about just setting the static variable holding the class in the NiftyDefaults class directly from the new method you've added to the Nifty class?

The new method can be none-static too (just so it can be used like all the other methods) and still sets the static variable in the NiftyDefaults class. We should note in the JavaDoc for that method that the user changes a (Nifty) system-wide setting though.

Besides that I'm happy with this pull request and that you've stored this setting in the NiftyDefaults class - where it belongs! :smile:

PS: What confused me a bit is that you've changed unrelated commets in the Nifty class. But I guess maybe your IDE did that automatically. I'm not against cleaning up such things. It just made it harder to actually see your important changes. Maybe we could separate this in different commits? Actual change and cleanups?

bgroenks96:
@void256 Yes, that sounds fine.

It looks like my IDE just stripped out extra whitespace characters at the end of some of the lines and added a few missing newlines.

I guess I can try to go separate that out if you really want... but that seems ilke a lot of effort for something so trivial :cold_sweat:

void256:
But you want to change your pull request a bit anyway, right? :smile: so you could clean up the comments as well :smile: and in general it's good style anyway. I mean, only change the necessary things in a single commit. Please? Pretty please? :sweat_smile:

bgroenks96:
Since you asked nicely....

bgroenks96:
> The new method can be none-static too (just so it can be used like all the other methods)

I didn't do this because it seemed way too deceptive to me... especially for all those programmers who barely read method names let alone Javadoc.

I think it needs to be kept static to make it clear that it is in fact a static property, not something that can be set per-Nifty instance.

bgroenks96:
Ahem.... @void256 >_>

void256:
Oops @bgroenks96 didn't saw the update. Thanks! Looks better now and has been applied :smile:

PR #360, opened 2015-08-11
The events should not be sent to Slick InputListener if the listener is not accepting the input.

Genhis:
You should consider that Slick InputListener doesn't have to want accept the events. In that case you should not forward those events to Slick.

That way is used in `org.newdawn.slick.Input#poll()`.


ghost:
You should put the "return false" on the same line as the "if(!listener.isAcceptingInput())".

Genhis:
Corrected. How should I update the pull request? Amend the commit or post new? This is my first pull request, so please advise me. :-)

bgroenks96:
For a small change like that, you can go ahead and just amend + push --force.

bgroenks96:
:+1:  Thanks!

PR #372, opened 2015-09-06
Layout fixes and reactivated use case

mkaring:


mkaring:
Already merged by a rebase.

PR #379, opened 2015-09-30
Fixed regression introduced by issue fix #118

BorealFeast:
Fixed regression introduced by issue fix \"Color field added to image element. #118\" where a colored image can no longer use the alpha value of the element.


bgroenks96:
Wow that's a pretty old regression bug that's gone unnoticed.... nice work spotting it, and thanks for the fix! Looks good :+1:

PR #390, opened 2016-01-01
Bugfix in Shader, Usage of Core Logger

mkaring:


void256:
Do we really want to use that CoreLogger class for this - or all over the place in the end? Really? o_O

IMHO the lib is called CoreGL and not CoreLoggingLib :yum: I mean it's only existence is to provide some help in using the GL. I'm not sure it's such a good idea to use it's Logger-class as well. Right now I still see that CoreLogger class as private API to ease logging inside of CoreGL and not really part of its public API.

In this specific case I'd vote to remove the logging line instead of adding a dependency to the CoreLogger class.

mkaring:
Well that is something to be decided. The core logger provides a cleaner way to write log messages like this and is a light weight helper class for logging. It can't be used in case there is no dependency to CoreGL, but since the dependency is there it is okay. But if you want I remove the commit from the pull request again.

void256:
I think it's not a good idea to log differently in the GL renderer so yes, please remove it.

PS: Maybe we (I mean I :wink:) should reconsider the usage of slf4j then. It would add an additional dependency of some 33 kb of memory footprint but we could use it everywhere in Nifty - which I'd prefer compared to having each module log differently.
PPS: On the other hand :wink: Nifty probably logs too much anyway. I'm not sure all of that logging is necessary at all. Argh, I'm still undecided if it's worth it to change to slf4j .. need to think about it some more and maybe we should discuss this somewhere else.

mkaring:
There you go

bgroenks96:
If you can stomach adding the dependency for SLF4J, I would go with that for Nifty. CoreLogger is really supposed to be for use by other CoreGL classes (and API users if they are keeping their project lightweight I suppose).

That being said, CoreLogger is a nice alternative if you don't want to add the dependency, as Martin said.

PR #396, opened 2016-01-12
fix: Sampler state 

dustContributor:
Sampler state gets stored, then sampler gets unbound so it doesn't overrides per-texture filtering, then it gets restored.

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/render/batch/core/CoreProfileSaveGLState.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/render/batch/core/CoreProfileSaveGLState.java
@@ -35,34 +35,49 @@ public class CoreProfileSaveGLState {
   }
 
   public void saveCore() {
+  	// Store currently bound program.
     params.clear();
     gl.glGetIntegerv(gl.GL_CURRENT_PROGRAM(), params);
     currentProgram = params.get(0);
 
+    // Store currently active texture unit.
     params.clear();
     gl.glGetIntegerv(gl.GL_ACTIVE_TEXTURE(), params);
     activeTexture = params.get(0);
 
+    // This texture unit will be used by Nifty.
+    gl.glActiveTexture(gl.GL_TEXTURE0());
+
+    // Store sampler bound to it.
     params.clear();
     gl.glGetIntegerv(gl.GL_SAMPLER_BINDING(), params);
     samplerBindingTex0 = params.get(0);
+    
+    // Now unbind sampler so it doesn't overrides texture filtering.
+    gl.glBindSampler(0, 0);
 
+    // Store texture bound to 2D target.
     params.clear();
     gl.glGetIntegerv(gl.GL_TEXTURE_BINDING_2D(), params);
     textureBinding = params.get(0);
 
+    // Store blending state.
     blending = gl.glIsEnabled(gl.GL_BLEND());
 
+    // Store blending source parameter.
     params.clear();
     gl.glGetIntegerv(gl.GL_BLEND_SRC(), params);
     blendingSrcFactor = params.get(0);
 
+    // Store blending destination parameter.
     params.clear();
     gl.glGetIntegerv(gl.GL_BLEND_DST(), params);
     blendingDstFactor = params.get(0);
 
+    // Store primitive restart state.
     primitiveRestart = gl.glIsEnabled(gl.GL_PRIMITIVE_RESTART());
 
+    // Store primitive restart index.
     params.clear();
     gl.glGetIntegerv(gl.GL_PRIMITIVE_RESTART_INDEX(), params);
     primitiveRestartIndex = params.get(0);
@@ -71,14 +86,23 @@ public class CoreProfileSaveGLState {
   }
 
   public void restoreCore() {
+  	// Restore program.
     gl.glUseProgram(currentProgram);
-    gl.glActiveTexture(activeTexture);
+    // Restore sampler that was bound to texture unit 0.
     gl.glBindSampler(0, samplerBindingTex0);
+    // Restore active texture unit.
+    gl.glActiveTexture(activeTexture);
+    // Restore bound texture to 2D target.
     gl.glBindTexture(gl.GL_TEXTURE_2D(), textureBinding);
+    // Restore blending state.
     enable(gl.GL_BLEND(), blending);
+    // Restore blending functions.
     gl.glBlendFunc(blendingSrcFactor, blendingDstFactor);
+    // Restore primitive restart state.
     enable(gl.GL_PRIMITIVE_RESTART(), primitiveRestart);
+    // Restore primitive restart index.
     gl.glPrimitiveRestartIndex(primitiveRestartIndex);
+
     CheckGL.checkGLError(gl, "Failed to restore OpenGL Core Profile state!", true);
   }
 


bgroenks96:
Nooooo. @dustContributor did not read all of my changes. Please see 9cfaa7e90b4 for further comments on this.

void256:
And see 70c4ffc1d296cc8a0fd660880511d457e8c8f3c2 as well :wink:

Should be fine now? Everyone happy? :smile:

dustContributor:
I **did** see all the changes, its just that binding it in the constructor would accomplish nothing. Now with the sampler binding in beginFrame its better.

Still, as I already mentioned, even with the sampler unbinding the user will need to call glActiveTexture(GL_TEXTURE0) before rendering nifty, for the reasons I listed in 9cfaa7e90b4255a664f8c6b2c5d5f849cfca7b8e 

Please, read it, I already described in detail how the sequence should work in the original issue and in that comment.

Otherwise users need to look at Nifty's sources to see which tex unit it will use, to bind it before rendering Nifty so the state tracker works properly. If you're not going to change that behavior, it should be documented.

bgroenks96:
@dustContributor  Ahem.... see [here](https://github.com/nifty-gui/nifty-gui/blob/1.4/nifty-core/src/main/java/de/lessvoid/nifty/render/batch/core/BatchRenderBackendCoreProfileInternal.java#L559)

void256:
@dustContributor that's correct and is still missing indeed. thank you!
@bgroenks96 the problem is that "glGetIntegerv(gl.GL_SAMPLER_BINDING(), params);" returns the sampler bound to the current active texture unit. We don't know which one it is in that save() method. So after saving the current texture unit we have to activate texture unit 0 before getting the GL_SAMPLER_BINDING. Otherwise we would reset the sampler to the wrong texture unit in restore which, in fact, right now we're doing!

Which is not correct!

PS: I did like the comments of dustContributor in the original pull request too. Any chance we can add these as well? It would further explain what's happening there.

dustContributor:
Praise the spaghetti monster!

bgroenks96:
Ah you meant for _reading_ the sampler state. Your comment seemed to be talking about binding texture 0 for rendering Nifty, i.e. binding it for Nifty to render to, which NIfty does do in the beginRendering method.

Yes at present it is possible for it to read the wrong sampler binding, so I guess we will need to add a call to activate the texture before reading it.... I'm not a huge fan of screwing around with state changes in the `saveState` method... but I think it should be ok for this.

Apologies for the confusion.

bgroenks96:
Fixed in ee5c708

R'Amen.

dustContributor:
Great :smile:

bgroenks96:
@void256 @dustContributor 

I'm still bothered by mutating state in the `saveState` method... and by the kind of hackiness of this fix. We're kind of assuming that Nifty will only use texture unit 0 with the default sampler. We should probably make it a TODO to integrate texture/sampler binding more robustly into the batch renderer.

It's a damn shame that Khronos hasn't included a mechanism in the spec for fetching the sampler for a non-active texture unit. Seems odd considering that you can bind it for a non-active texture unit...

dustContributor:
Leeeave it to the user, leeeeave it to the uuuuser...

I insist Nifty can't possibly account for all the weird state combinations OpenGL can have. Trying to track the state on its own bubble is a futile effort. Eventually someone else will come with a XYZ state configuration that Nifty didn't accounted for, and once all the cases are covered someone will come saying "the state tracker is too slow!".

Its just easier for everyone involved to make a list of how Nifty expects OpenGL state will be, and a list of how it will leave the state once its finished, kinda like NanoVG does.

In my case I think I'll replace that CoreProfileSaveGLState instance with a subclass that does nothing (with reflection since its a private field of the batch renderer) then track state changes from overriden methods in CoreGL subclass... As long as you don't make the classes/methods final I can work around it.

void256:
I agree with @dustContributor especially after what he explained in the original issue as well (stalling the pipeline with all of our glGet*) - btw thanks for the explanation @dustContributor 

Maybe we should make the CoreProfileSaveGLState class available as a somekind of utility class that the user can call when he wants to or sees the need and otherwise require the user to set GL state correctly for Nifty rendering himself.

Thinking of that it would be possible to make another class available to setup GL state for Nifty rendering  which can act as documentation as well :smile:

We're going to definetely consider this for 2.0 as well!

bgroenks96:
@dustContributor If you need to use reflection to do something like that in Nifty, that would be an indicator of an API failure on Nifty's part. If you have an idea for how to make CoreProfileSaveGLState more pluggable/extensible, please submit a PR (or let us know what we could change). That would be really helpful.

@void256 I'm not talking about additional `glGet` calls. I'm just talking about not hiding sneaky assumptions into discrete parts of the codebase like choosing the sampler texture unit (and setting the active texture) in the `saveState` method. That just seems like future bugs and hair pulling waiting to happen.

I agree that `glGet` is bad. IMHO each instance of CoreProfileSaveGLState should just have a number of state parameters (immutable or mutable) configured by the user for their needs (or by Nifty internally for its needs) and then simply set those values on a call to `restoreState` rather than try to be smart and query them from the GL driver.

void256:
@bgroenks96 I know what you mean :wink: and I in general I agree that its not optimal to change GL state in that save() method. However, in this very specific case (glActiveTexture) I don't mind. We don't do any harm to any GL state. The only real drawback I see is that we call glActiveTexture one time too much but that's it.

And in general I agree with dustContributor .. the user has to set the required state before calling nifty.render(). That's the best way from a performance perspective. We just need a proper way to document our requirements.

And as I said before we could always provide some form of helper class to do that. But using it or not would be totally up to the user.

bgroenks96:
@void256 I'm not sure I agree that the user should be responsible for setting Nifty's state requirements. That seems odd. Nifty should take care of setting the state according to its own requirements.

Restoring the state, however, could be a user responsibility, and we could provide a helper class (similar to CoreProfileSaveGLState) that could hold user configured values (not values from glGet) that would be set on a call to `restoreState`.

void256:
As dustContributor wrote on the original issue (and here as well) telling the user what GL state Nifty requires and letting him set it before calling nifty.render() **is** going to be the best way (performance wise).

It simply reduces state changes to the minimum. If some GL state is already set (because the user needs it too) there is no need to save the state or set it again or call some Nifty class to set some state variables. He simply can let everything as it is and call Nifty. Done.

bgroenks96:
I have never seen another GL application/library ask its user to set all of the appropriate states before calling it. That's just confusing. If you're going to do that, it needs to be extremely well documented. i.e. not just a single JavaDoc comment. It should be made immediately clear to someone as soon as they start using Nifty that they are going to have to make those state change calls, else they might have a ton of hair pulling right out of the gate and just give up on it.

Ideally, we would have a GL interface layer (cough cough.... CoreGL) that would intercept redundant state change calls. Then this would be a non-issue.

dustContributor:
> I have never seen another GL application/library ask its user to set all of the appropriate states before calling it.

First time for everything :smile: I suppose you already looked at NanoVG's GitHub, there it just lists the state it will touch for example, libRocket goes one step beyond and simply makes a renderer interface that the user must implement to get anything drawn http://librocket.com/wiki/documentation/C%2B%2BManual/Interfaces

>  If you have an idea for how to make CoreProfileSaveGLState more pluggable/extensible

I guess there could be some sort of GLStateManager interface, with one implementation that does everything in its own like the current does (glGet and all), other has pluggable state fields that the user can set, ie:

```
GLStateManager tracker = new TrackingGLStateManager();
// So these won't be queried with glGet
tracker.setActiveTexUnit(myActiveTexUnit);
tracker.setBoundVertexArray(myBoundVAO);
// etc
```

While the tracker tracks only the state Nifty modifies. Rest of the state is up to the user to make it so its correct (stencil, glDrawBuffers, etc).

Nifty sets which texture units the manager has to track by passing an int[] with their indices to it. ie "stateTracker.trackTextureUnits(new int[]{0, 1, 5});" Inside it would create a new array for holding the bound texture to those units for later restoring. If an user wants no tracking, it can provide some "NullGLStateManager" implementation that does nothing.

In any case you go back to the first issue, in reality even if Nifty doesn't touches say, stenciling, Nifty needs it to be turned off. Even if Nifty doesn't touches a framebuffer, it needs some draw buffer selected, and so on.

Since OpenGL is a state machine, Nifty implicitly needs all of it to be set in a particular way plus it needs to modify a couple more states explicitly. The implicit state leads to hard to track bugs, like the one with the sampler object I had.

In my renderer I ended up specifying all the state I track for all the passes because leaving state not explicitly defined leaded to "state bleeding" in between passes (ie, one pass didnt touched blending but what it really needed was blending disabled, previous pass touched blending and now you got a hard to track bug that manifests itself in a render pass that isn't even the issue).

bgroenks96:
> Since OpenGL is a state machine, Nifty implicitly needs all of it to be set in a particular way plus it needs to modify a couple more states explicitly.

Yep. That's why it's usually a good idea to reset states after you change them, especially for GL_STENCIL_TEST and GL_DEPTH_TEST.

The GLStateManager idea seems workable. I like the idea of having pluggable implementations (and I'm always in favor of dependency injection).

PR #399, opened 2016-01-20
Updated Processing Renderer and Minim Soundsystem to Processing 3.0.1

barefists:
Updated the renderer, input and soundsystem to work properly in the new Processing.

I've also ported over the Nifty examples to Processing's .pde format. Since these aren't exactly native Java files, how should I go about adding them to the project? Also, they use .jar versions of nifty, so should I include the .jars as well in the commit?


bgroenks96:
Could you do a `git rebase -i` and squash the WIP and merge commit? Would prefer not to have only locally relevant commits polluting the main branch commit history, if possible.

As for the .pde thing, I don't really know what that is :cold_sweat: ... would users of the nifty Processing renderer need them in order to run examples? If they are binaries, ideally we would be able to construct them in the build rather than keep them in the repository. Then again, I know how much Maven sucks to customize with stuff like that, so I wouldn't ask you to integrate it in if it's too complicated.

Other than that, I don't think we have much in the way of tests/examples currently for this renderer. So assuming you've tested this with some code using the renderer and everything works, I'm inclined to just merge it :wink:

barefists:
Ah, I've gone ahead and messed up my fork due to my noobness in git trying to rebase it. Can you reject this pull request? I'll paste my local backup of the changes into a new fork and submit that instead.

As for .pde files, they are source files used in Processing's own IDE, written in the Processing language (which is like a less verbose dialect of Java). I'm not sure how to fit it into Maven, so maybe I'll just maintain it separately from Nifty, and let it live its life on the Processing forums.

bgroenks96:
If they are source files, you can go ahead and commit them in. Just make a new examples module.

And ok, I will be awaiting your new pull request!

Issue #391, opened 2016-01-01
NiftyColor


mkaring:
The NiftyColor class needs some reconsideration in the current implementation for Nifty 2.0

The functions for fetching the "standard" colors do not contain any caching of instances but instead perform a expensive parsing from a string constance every time the function is called.

What was the idea behind implementing those constant colors as functions and not as static constants? The `NiftyColor` class is immutable and would work very well when used as constants?

For our examples this class is currently mostly responsible for the generated memory garbage.

void256:
I agree! This should be changed. The static methods should return color constants instead of parsing the color strings on each call. That sounds indeed really stupid in retrospective. Preventing memory garbage is essential for Nifty!

I don't see any real reason that these are static methods instead of constants. However, having them as methods will eventually provide a better API evolution compared to allowing direct access to a static constant. That's the only idea that might have motivated it.

relu91:
> I don't see any real reason that these are static methods instead of constants. However, having them as methods will eventually provide a better API evolution compared to allowing direct access to a static constant. That's the only idea that might have motivated it.

So what should we choose? public static methods with private static variables or only public static variables ? 
I can handle this! ^^

relu91:
There's some reason to leave static methods there? maybe they are called somewhere using reflection?

void256:
I would suggest to just keep the public static methods and return a private constant.

relu91:
Why? I know that NiftyColor is widely used in nifty and a change to the api means a lot of changes somewhere else, but I don't think it's worth to keep static methods, it only introduces overhead...

void256:
Because I like the colors to be written in lower case letters and if they are constants we would probably write them in upper case? :wink:
I don't think it's going to be a catastrophic runtime cost to have these as methods instead of constants. And if we ever want to track something silly like how many times a specific color is being used .. then we only can do that with methods and not when we reference the constants directly :smile:

bgroenks96:
I agree with @void256 about returning the private constant. The constants can be initialized at load time in the same manner the public methods are doing it now, and then the public methods can just return those values. Hooray for efficiency!

There are other less silly reasons to do it this way as well. For example, we may at some point want to allow manipulation of NiftyColor so that one or more of the color channels is always biased by some pre-configured factor (i.e. maybe having all colors offset by a gamma constant).

void256:
[...]

PR #403, opened 2016-01-24
NiftyColor refactor

relu91:
As we discussed in #391 I changed accordingly the NiftyColor class. I've not added the color instance at methods invocation time, maybe we could add it later if there's some memory space problem. 

P.s. I've added also a style commit using the 1.4 nifty-gui formatter.


void256:
I was about to complain about the style changes ... but then realised there are two commits! :smile:
Looks good to me! Thanks! :smile:

PR #405, opened 2016-01-25
Modified ClassHelper so that the context classloader is used.

shamanDevel:
I'm using Nifty inside the netbeans platform. There, the 'normal' classloader can only load classes from dependencies.
This makes it impossible to register new ScreenControllers or effects via XML, they simply cannot be found.
This is fixed by using the context classloader instead which has access to all classes.
Outside the netbeans platform, the context classloader is just the normal classloader, so the functionality is also preserved in normal scenarios.


bgroenks96:
Looks good. Thanks!

:+1:

PR #407, opened 2016-01-30
Upgrade examples with new layout apis 

relu91:
I fix some of the examples upgrading the code with new layout api interface made by @mkaring. I've done what it could be possible and to the rest that still need the fix I added some comments to explain the issue. I hope this could be useful in the future!


void256:
Great work @relu91! Thanks! :smile:

PR #408, opened 2016-01-31
fix: Fixed tests for both layout node implementations

mkaring:
The tests were not updated for the new layout implementations. Now they work again.


void256:
Thanks! :smile:

And I've finally figured out how to integrate changes without an additional merge commit. I've documented it in the wiki: https://github.com/nifty-gui/nifty-gui/wiki/Collaborators-and-Team-Member-advice

PR #418, opened 2016-03-20
 Utility classes should not have public constructors

AymanDF:
This pull request is focused on resolving occurrences of Sonar rule squid:S1118 - “Utility classes should not have public constructors ”. You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S1118
Please let me know if you have any questions.
Ayman Abdelghany.


void256:
Thanks for your efforts Ayman!
I do understand that rule. However, I don't think I really agree to applying it blindly to all the UseCase_\* classes. They are meant to be examples and I don't think we'd really benefit of applying a private constructor to these classes. It would not hurt to add it but there really is no value in it for these specific classes and I do hesitate a bit to add them there.

bgroenks96:
Modified and merged with 6f79f8b

PR #426, opened 2016-05-30
Add nifty-renderer-lwjgl3 module.

### LINKED ISSUE #304: LWJGL 3 support.
Hi! LWJGL 3 has been released. Most (if not all) the GLxx.xx calls are exactly the same but there are a few things here and there that changed (GLU is gone IIRC).

bgroenks96:
Initial (untested!) port of Nifty 1.4.x renderer from LWJGL2.x -> LWJGL3

The only noticeable difference from LWJGL nifty components is the requirement for a
GLFW window handle, which the user _should_ have if using LWJGL3.

Additional changes:
- Copy nifty-examples-lwjgl -> nifty-examples-lwjgl3
  Example runner has not yet been implemented for LWJGL3, so these don't work yet.

Resolves #304


bgroenks96:
If someone familiar with setting up LWJGL3 applications could go ahead and test this before I get around to it, I would much appreciate it!

Guerra24:
Hello, i tested with my game and worked good, the only problem i see is that the mouse/keyboard callbacks broken my mouse/keyboard system. So i recommend leaving that to discretion of the user but apart from that everything works.

bgroenks96:
@Guerra24 Hmm. Nifty needs input callbacks somehow.... a bit annoying the LWJGL3 uses GLFW's single callback mechanism directly without any abstraction layer that allows for multiple listeners.

Do you have a preference from an API user's point of view how you would like this to be setup? Perhaps you call a Nifty method and supply the callback arguments?

bgroenks96:
@Guerra24 See if 7f7eb47 works for you. You should be able to invoke the Nifty input callback handlers from your own. You'll just have to keep a reference to the input system around.

Guerra24:
Now it's fully working, there is no problems with rendering or input.  Thanks @bgroenks96 for the LWJGL3 version :D.

bgroenks96:
No problem, @Guerra24. Thanks for testing it. I'll go ahead and merge this into the main branch and talk to @void256 about shipping it in 1.4.3 somewhat soon.

PR #428, opened 2016-06-08
fix ibxm.jar dependency,

jjYBdx4IL:
replace strange nifty-version property with project.version


mkaring:
I don't think 8312acf will be accepted. We avoided to this point to have any IDE specific configuration inside the repository.

jjYBdx4IL:
removed.

PR #453, opened 2018-07-29
Added more interact methods to ElementBuilder to match XML format

Genhis:
I've added other interact methods to `ElementBuilder` to match XML format and reordered the old ones, so that it matches the order in `InteractType.java` and `nifty.xsd`.


void256:
Looks good to me :smile: Thanks!

PR #454, opened 2018-07-31
keyboard effects not being run fix

joliver82:
Add call to run the onclick effect from clickAndRelease method so it's run when keyboard activation events.

Should fix issue https://github.com/nifty-gui/nifty-gui/issues/452


Genhis:
@joliver82 @void256 This commit needs to be either fixed or reverted. Eclipse is complaining about missing `EffectEventId` import and although Maven initially returned no error, Java Decompiler showed this:

```
  public void clickAndReleaseMouse(@Nonnull Nifty paramNifty) {
    throw new Error("Unresolved compilation problem: \n\tEffectEventId cannot be resolved to a variable\n");
  }
```

void256:
There was an import missing in this merge request. I fixed this with ae80a39998881ca20f268884ef34988ff2ea00ac

Issue #41, opened 2012-08-31
DSLification


toolforger:
I'm currently preparing a pull request that has modified `*Builder` classes that allow stuff like
    `new ButtonBuilder().widthPx(102).text("Cancel").build(nifty, screenId, parent)`
Existing classes still have an identical API, with the following exceptions:
- All builders have a parameterless constructor now, so people can do `new ButtonBuilder().id("someId")`. This allows using builder-creating functions to generate a control template that can have its id filled in by the caller, i.e. it decouples builder construction from id assignment. (It also plays better with frameworks that use reflection, though I don't have a use case for that.)
- The `build()` functions return the control created. I.e. `ButtonBuilder#build()` returns a `Button`, etc. The obvious exceptions being `LayerBuilder`, `PanelBuilder` etc, which return just the `Element`.
- Possibly remaining silly mistakes.

I'd be happy to hear what you think about that.

I also have some further plans and would appreciate feedback:
1) I'd like to make a `build(parent)` function that picks up the Nifty and the screenId from parent. For that purpose, would it be okay to add a `getScreen()` to Element, and a `getNifty()` to screen?
2) I'd like to make `onShow()`, `onClick()`, ..., etc. functions in `ElementBuilder` and subclasses. The parameter should be something executable (i.e. a Runnable or an EventTopicSubscriber). This would allow writing DSL code like this (not tested):

``` java
Panel parent = ... .build();
TextField characterName = new TextFieldBuilder(). ... .build(parent);
Button button =
  new ButtonBuilder().widthPx(150).text("OK").onClick(
    new Runnable(){public void run() {
      gameCharacter.setName(characterName.getRealText());
    }}
  ).build(parent);
```

What's the best API that the onXxx routines should use? Set up an `EventTopicSubscriber` (that would have to be postponed until build() time I guess), or is that already available somewhere? I have seen the `interactOnXxx` functions, but Strings can contain typos and aren't properly updated during refactoring, so I don't want to use these.

void256:
Thanks for your effort. This looks pretty good.

Just make sure to make these changes in the 1.4 branch. Even though they seem to be pretty compatible it's probably too late for 1.3.2.

1) Adding getScreen() and getNifty() sounds ok to me.

2) I don't think that's going to work without changing a lot. The interact onClick() and so on methods are all meant to be called on a controller or ScreenController. So maybe that's more like something for 2.0? ;)

toolforger:
[...]

PR #456, opened 2018-08-09
Simpler Element building

Genhis:
`Element` objects already hold `Nifty` and `Screen` instances, so we can use them when building new elements which would result in simpler API for the Nifty user.

@void256 @bgroenks96 Although I am not using most of the default controls, I can do the refactoring proposed in #41. That is to return `this` instance in setter methods for stacking (without event functions such as `onClick`, etc.). I remember this was the thing which discouraged me from using Nifty when I first started. Shall I do the refactoring?

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/builder/ElementBuilder.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/builder/ElementBuilder.java
@@ -489,9 +489,9 @@ public abstract class ElementBuilder {
    *
    * @return the element created
    */
-  public Element build(@Nonnull final Nifty nifty, @Nonnull final Screen screen, @Nonnull final Element parent) {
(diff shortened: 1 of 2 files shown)


void256:
Merged, Thanks :smile:

Feel free to do the refactoring. I don't see that returning this is breaking anything and if it makes usage of the Builders more easy then yes, go for it 👍

void256:
This merge request broke the unit test CreateCheckBoxControlTest. I'm fixing it.

void256:
ControlsDemoMain in the nifty-examples-lwjgl project failes with a NPE as well 😞

void256:
The CreateCheckBoxControlTest is fixed with the new commit 32911e3a10b7352c4e9cee8395872c7752d959be. However, the NPE still exists.

This should be fixed in merge request #457 as well.

PR #466, opened 2019-05-25
fixes custom parameter resolving for nested controls

### LINKED ISSUE #465: Parameter not resolved correctly for nested controls
Currently, if you define nested controls with different parameters, the parameter is resolved with the value declared on the outer most control. This effectively prevents the use of individual parameter for nested controls, which is not very nice if one wants to create some kind of tree structure (like a proper menu). Debugging shows that de.lessvoid.nifty.loaderv2.types.ElementType#resolveParameters passes the attributes down to child elements and merges them with existing attributes. If the attribute is already present, it is silently rejected. Wouldn't it be better if new declarations override existing ones,  so that the described scenario would work? Would that break something I´m not aware of? I´m also happy to submit a pull request if the change is desired.

Tasemo:
Fixes #465 by overriding existing attributes when resolving them, so that nested controls can use individual parameters. All tests are green and I have run the AllControlsDemo and did not find any problems. Since this is my first commit on this project, feel free to point out any mistake or violation of code conventions you guys might have.

### DIFF
--- a/nifty-core/src/test/java/de/lessvoid/nifty/loaderv2/NiftyLoaderTest.java
+++ b/nifty-core/src/test/java/de/lessvoid/nifty/loaderv2/NiftyLoaderTest.java
@@ -1,6 +1,8 @@
 package de.lessvoid.nifty.loaderv2;
 
 import de.lessvoid.nifty.Nifty;
+import de.lessvoid.nifty.elements.Element;
+import de.lessvoid.nifty.elements.render.TextRenderer;
 import de.lessvoid.nifty.screen.Screen;
 import de.lessvoid.nifty.screen.ScreenController;
 import de.lessvoid.nifty.spi.input.InputSystem;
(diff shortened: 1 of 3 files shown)


nifty-core/src/test/java/de/lessvoid/nifty/loaderv2/NiftyLoaderTest.java
```diff
@@ -1,20 +1,30 @@
...
+import org.junit.rules.TemporaryFolder;
 
 import javax.annotation.Nonnull;
 import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.InputStream;
+import java.io.PrintWriter;
 
-public class NiftyLoaderTest extends TestCase {
+import static junit.framework.Assert.assertEquals;
```
Genhis, line 25: This class is deprecated, use `org.junit.Assert.assertEquals` instead?

Genhis (approved):
Looks good and +1 for writing tests :-)

All unit tests have passed, and I have run nifty-examples with the new code which all seem to be working correctly.

@void256 @bgroenks96 Can it be merged, please?

bgroenks96 (approved):
Looks good, thanks!

PR #468, opened 2019-06-11
introduces minimum size value

Tasemo:
Hey guys,
I had the need for a new size value that sets the size of an element to the minimum size of all children. While I understand that this makes not too much sense while looking at average use cases (why would someone make a container not big enough to hold all children), I don´t think it would do any harm to add it. For example see the following example where the texts overlap, but only one text is visible at a time: 
```xml
<panel childLayout="horizontal">
	<panel childLayout="vertical" width="min">
		<text text="Menu1" font="aurulent-sans-16.fnt"/>
		<text text="LongerTextThatIsNotAlwaysVisible" font="aurulent-sans-16.fnt" visible="false"/>
	</panel>
	<panel childLayout="vertical" width="min">
		<text text="Menu2" font="aurulent-sans-16.fnt"/>
		<text text="LongerTextThatIsNotAlwaysVisible" font="aurulent-sans-16.fnt" visible="false"/>
	</panel>
</panel>
```
Is there any other way of achieving the same structure ("Menu1" and "Menu" next to each other, the longer texts beneath each menu entries) without using absolute values?

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/elements/Element.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/elements/Element.java
@@ -769,7 +769,7 @@ public class Element implements NiftyEvent, EffectManager.Notify {
 
     if (layoutManager != null) {
 
-      // unset width, or width="sum" or width="max"
+      // unset width, or width="sum", width="max" or width="min"
       // try to calculate the width constraint using the children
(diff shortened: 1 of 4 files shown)


Genhis:
Can you post a screenshot of the intended use-case and another one to see how it looks without your changes? There may be another way to do it.

PR #470, opened 2019-06-14
sets visibleToMouse to true for onClick effects

### LINKED ISSUE #469: visibleToMouse not set automatically when onClick effect is added
Hey guys,
I noticed that the visibleToMouse attribute is automatically set to true when a \<interact> tag or an \<onHover> effect is attached to an element. This is not the case for the \<onClick> effect, even though this is always required to make the effect work. The following examples work and set the attribute automatically:
``` xml
<panel childLayout="center">
    <effect>
        <onHover name="nop" onEndEffect="effectExecuted()"/>
    </effect>
</panel>
<panel childLayout="center">
    <interact onClick="effectExecuted()"/>
</panel>
```
The following example does not work without setting visibleToMouse to true:
``` xml
<panel childLayout="center">
    <effect>
        <onClick name="nop" onEndEffect="effectExecuted()"/>
    </effect>
</panel>
```

Tasemo:
Fixes #469. Not much else to say about this.

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/effects/Effect.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/effects/Effect.java
@@ -100,8 +100,7 @@ public class Effect {
   public void enableHover(final Falloff falloffParameter) {
     hoverEffect = true;
     falloff = falloffParameter;
-
-    element.setVisibleToMouseEvents(true);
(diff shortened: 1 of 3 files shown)


bgroenks96:
Looks good, thanks. Sorry for not merging this before.

PR #487, opened 2021-02-04
Instead of subscribing to the EventService, refresh styles by travers…

### LINKED ISSUE #461: Memory Leak in style event service
So I had a few users complain about a performance problem in my game in an area that was adding and removing a lot of nifty elements. After looking into it I found the following call in element construction with no corresponding call when removed:

`nifty.getEventService().subscribeStrongly("style-refresh:" + getStyle(), styleListener);`

This method basically loops through every single element that is subscribed with the same style name.  After adding and removing elements after an hour+ of gameplay this causes some pretty major performance issues. This listener doesn't appear to be used unless you are trying to change a style by hand since I removed it in a local nifty build and haven't had any negative effects. 

Is there a reason to have a subscription service to update styles as opposed to just walking the scene graph? It seems like a hefty performance hit for adding items in order to update styles.

tonihele:
…ing the scene graph

This fixes the slowdowns and the memory leaks caused by the subscriptions that are never used. I suspect this feature was made for the Nifty UI editor (WYSIWYG Nifty building). I bet the actual usage in applications is very rare. The mentioned editor doesn't even work with anything newer than Nifty 1.4.0. Well, the XML it produces does work, so in that sense it usable.

Resolves #461
https://github.com/nifty-gui/nifty-gui/issues/461

### DIFF
--- a/nifty-core/src/main/java/de/lessvoid/nifty/Nifty.java
+++ b/nifty-core/src/main/java/de/lessvoid/nifty/Nifty.java
@@ -1,34 +1,5 @@
 package de.lessvoid.nifty;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Properties;
-import java.util.ResourceBundle;
-import java.util.Set;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-import javax.annotation.WillClose;
-
-import org.bushe.swing.event.EventService;
-import org.bushe.swing.event.EventServiceExistsException;
-import org.bushe.swing.event.EventServiceLocator;
-import org.bushe.swing.event.EventTopicSubscriber;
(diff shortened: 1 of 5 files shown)


bgroenks96:
@void256 Any reason why we shouldn't merge this? I must have missed it.

bgroenks96:
Ok, I don't have a Java or nifty build environment set up at the moment, so can you @tonihele verify that this branch still builds? Maybe post a screenshot or something?

tonihele:
I use a fork with this merged and there are also other forks that have merged this. I know that Nifty is no longer actively maintained, but I really suggest https://docs.github.com/en/actions/guides/building-and-testing-java-with-maven. I did this to my repo (OpenKeeper, not the Nifty fork.. I should maybe put to there also) just recently, just really one or two clicks. And then at least the repo has a basic CI in place, to see if PRs and master builds.

If you really need the screenshot, let me see in the evening if I can get one.

relu91 (approved):
For what it's worth I reviewed the changes trying to remember why it was introduced. Quoting @tonihele, it was indeed introduced to update a style dynamically in the Editor. One use case for games that I remembered was for previewing changes in the style of the GUI (i.e. mods). The change seems to keep the support of the feature but probably making it a little bit slower but since it is a corner case I totally support it!

bgroenks96:
@tonihele Ok, I think that's fine. Nifty had automated CI at one point, I think (provided by @mkaring ?), but it seems to not be around anymore. GitHub pipelines would indeed be nice to add, but I don't have time at the moment to set this up.

